PR #84, opened 2022-05-25
Fix ResourceStatus serialization bug

jasongin:
- `ResourceStatus` serialization/deserialization could infinitely recurse in some cases. There's not a simple way to ask the serializer to do default serialization (without the converter to avoid recursing), so it takes a little more code to handle it explicitly while respecting other serializer options.
 - After reviewing other converter classes, I updated the `TunnelAccessControl` converter to better respect serializer options.
 - Implicit conversions between `ResourceStatus` and `ulong` simplify status-reporting and testing code.

### DIFF
--- a/cs/src/Contracts/ResourceStatus.cs
+++ b/cs/src/Contracts/ResourceStatus.cs
@@ -32,17 +32,24 @@ public class ResourceStatus
     public ulong? Limit { get; set; }
 
     /// <summary>
+    /// Implicitly converts a number value to a resource status (with unspecified limit).
+    /// </summary>
+    /// <param name="value"></param>
+    public static implicit operator ResourceStatus(ulong value)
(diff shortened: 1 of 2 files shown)


IlyaBiryukov:
Are there unit tests?

IlyaBiryukov:
Do we need to also change other SDKs in go, java, ts?

IlyaBiryukov (approved):

jasongin:
> Are there unit tests?

Not here, but I wrote status-update tests in the service repo that cover this code (not yet shared in a PR), which is how I found the bug.

> Do we need to also change other SDKs in go, java, ts?

No. Go and Java custom serializers don't have this problem. TS doesn't really deserialize since the type is a representation of the actual JSON structure.

PR #86, opened 2022-05-27
Fix date properties after deserializing in TS SDK

jasongin:
TS contract interfaces declare some properties as `Date` type. I noticed they are actually getting deserialized as strings (which makes sense). There's no easy way to define custom deserializers in TS/JS. So I added some code to fix up those properties in each API that can return time. It does mean we'll have to update that code if we add more `Date` properties to any contracts.

### DIFF
--- a/ts/src/management/tunnelManagementHttpClient.ts
+++ b/ts/src/management/tunnelManagementHttpClient.ts
@@ -30,6 +30,32 @@ function comparePorts(a: TunnelPort, b: TunnelPort) {
     return (a.portNumber ?? Number.MAX_SAFE_INTEGER) - (b.portNumber ?? Number.MAX_SAFE_INTEGER);
 }
 
+function parseDate(value?: string | Date) {
+    return typeof value === 'string' ? new Date(Date.parse(value)) : value;
+}
+
+/**
+ * Fixes Tunnel properties of type Date that were deserialized as strings.
+ */
+function parseTunnelDates(tunnel: Tunnel | null) {
+    if (!tunnel) return;
+    tunnel.created = parseDate(tunnel.created);
+    if (tunnel.status) {
+        tunnel.status.lastHostConnectionTime = parseDate(tunnel.status.lastHostConnectionTime);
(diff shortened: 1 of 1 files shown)


ts/src/management/tunnelManagementHttpClient.ts
```diff
@@ -30,6 +30,32 @@ function comparePorts(a: TunnelPort, b: TunnelPort) {
...
 }
 
+function parseDate(value?: string | Date) {
+    return typeof value === 'string' ? new Date(Date.parse(value)) : value;
+}
+
+/**
+ * Fixes Tunnel properties of type Date that were deserialized as strings.
+ */
+function parseTunnelDates(tunnel: Tunnel | null) {
```
jessetrinity, line 40: `tunnel?:`

ts/src/management/tunnelManagementHttpClient.ts
```diff
@@ -30,6 +30,32 @@ function comparePorts(a: TunnelPort, b: TunnelPort) {
...
+    if (tunnel.status) {
+        tunnel.status.lastHostConnectionTime = parseDate(tunnel.status.lastHostConnectionTime);
+        tunnel.status.lastClientConnectionTime = parseDate(tunnel.status.lastClientConnectionTime);
+    }
+}
+
+/**
+ * Fixes TunnelPort properties of type Date that were deserialized as strings.
+ */
+function parseTunnelPortDates(port: TunnelPort | null) {
```
jessetrinity, line 52: `port?:`

ts/src/management/tunnelManagementHttpClient.ts
jasongin: No, it should be an object or `null`, never `undefined`.

ts/src/management/tunnelManagementHttpClient.ts
jasongin: It should be an object or null, never undefined, because this is a result object that is being returned to the caller of the API.

jessetrinity (approved):

PR #88, opened 2022-06-10
Include auth params in unauthorized exception

jasongin:
Fix the C# tunnel management client to propagate additional parameters related to custom auth in case of an unauthorized error. (Other language SDKs don't currently have any special handling of the WWW-Authenticate response header.)

### DIFF
--- a/cs/src/Management/TunnelManagementClient.cs
+++ b/cs/src/Management/TunnelManagementClient.cs
@@ -403,7 +403,8 @@ namespace Microsoft.VsSaaS.TunnelService
                         if (response.Headers.WwwAuthenticate?.Count > 0)
                         {
                             ex.SetAuthenticationSchemes(
-                                response.Headers.WwwAuthenticate.Select((v) => v.Scheme).ToArray());
+                                response.Headers.WwwAuthenticate.Select(
+                                    (v) => v.ToString()).ToArray());
                         }
 
                         throw ex;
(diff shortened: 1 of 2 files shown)


cs/src/Management/TunnelManagementClient.cs
```diff
@@ -403,7 +403,8 @@ private static void ValidateHttpHandler(HttpMessageHandler httpHandler)
                         if (response.Headers.WwwAuthenticate?.Count > 0)
                         {
                             ex.SetAuthenticationSchemes(
-                                response.Headers.WwwAuthenticate.Select((v) => v.Scheme).ToArray());
+                                response.Headers.WwwAuthenticate.Select(
+                                    (v) => v.ToString()).ToArray());
```
IlyaBiryukov, line 407: Relay service uses this authentication schemes of `UnauthorizedAccessException` [here](https://devdiv.visualstudio.com/OnlineServices/_git/tunnels?path=/src/Tunnel/Src/DataPlane/Relay/ControlPlaneConnection/Exceptions/ControlPlaneUnauthorizedAccessException.cs&version=GBmain&line=28&lineEnd=29&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents)
Does it need to be updated?

jasongin: It will, but not immediately. Currently the control plane only returns the simple scheme name in the `WWW-Authenticate` response header, so this `.ToString()` is the same as `.Scheme` was before.

For domain custom auth, I am changing the control plane to return additional parameters here, and updating the relay service and CLI to handle the optional additional parameters.

jasongin:
I updated the `Get`/`Set` extension methods to use the structured `AuthenticationHeaderValue` class instead of `string`.

IlyaBiryukov (approved):

PR #89, opened 2022-06-11
Fix handling of multiple WWW-Authenticate header values

jasongin:
The tunnel service can potentially return multiple values for the `WWW-Authenticate` header. The multiple values were not getting propagated as expected by the SDK client due to a bug or limitation in .NET's `HttpResponseHeaders` class.

### DIFF
--- a/cs/src/Management/TunnelManagementClient.cs
+++ b/cs/src/Management/TunnelManagementClient.cs
@@ -400,11 +400,10 @@ namespace Microsoft.VsSaaS.TunnelService
                     case HttpStatusCode.Forbidden:
                         var ex = new UnauthorizedAccessException(errorMessage, hrex);
 
-                        if (response.Headers.WwwAuthenticate?.Count > 0)
-                        {
-                            ex.SetAuthenticationSchemes(response.Headers.WwwAuthenticate);
-                        }
-
+                        // The HttpResponseHeaders.WwwAuthenticate property does not correctly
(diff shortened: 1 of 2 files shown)


cs/src/Management/TunnelManagementClient.cs
```diff
@@ -400,11 +400,10 @@ private static void ValidateHttpHandler(HttpMessageHandler httpHandler)
...
                         var ex = new UnauthorizedAccessException(errorMessage, hrex);
 
-                        if (response.Headers.WwwAuthenticate?.Count > 0)
-                        {
-                            ex.SetAuthenticationSchemes(response.Headers.WwwAuthenticate);
-                        }
-
+                        // The HttpResponseHeaders.WwwAuthenticate property does not correctly
+                        // handle multiple values! Get the values by name instead.
+                        var authHeaderValues = response.Headers.GetValues("WWW-Authenticate");
```
IlyaBiryukov, line 405: Nit: it can be `HeaderNames.WWWAuthenticate`

IlyaBiryukov (approved):

cs/src/Management/TunnelManagementClient.cs
jasongin: `HeaderNames` is in the `Microsoft.Net.Http.Headers` assembly, which is not currently referenced by the SDK, and I don't think it's worth adding the dependency just for this.

PR #91, opened 2022-06-17
Initial working version of Rust SDK

connor4312:
(I merged main into my branch, which is why there are some extra commits on here)

This is an initial implementation of the management APIs in Rust SDK, based somewhat on the TypeScript SDK. It updates contract generators to work, though there are some things (as I mentioned in teams) that are plastered over for right now.

It also has some end to end tests -- the process of invoking these is fairly manual, but they're useful.


cs/tools/TunnelsSDK.Generator/RustContractWriter.cs
```diff
@@ -64,6 +80,44 @@ public override void WriteContract(ITypeSymbol type, ICollection<ITypeSymbol> al
         File.WriteAllText(filePath, s.ToString());
     }
 
+    public override void WriteCompleted()
+    {
+        this.WriteModRs();
+    }
+
+    private void WriteModRs()
```
jasongin, line 88: Great, I had been thinking it would be nice to generate this also, thanks for adding it.

cs/tools/TunnelsSDK.Generator/RustContractWriter.cs
```diff
@@ -309,18 +410,19 @@ private string GetRustTypeForCSType(string csType, IPropertySymbol property, Sor
...
         {
             rsType = $"Vec<{rsType}>";
         }
 
         if (isNullable)
         {
+            if (rsType == "String")
+            {
+                imports.Add("crate::contracts::serialization::empty_string_as_none");
+                s.AppendLine("    #[serde(default, deserialize_with = \"empty_string_as_none\")]");
```
jasongin, line 423: There are certain cases where the distinction between empty string and none/null is important. For example when updating a tunnel, a null description property indicates it should not be updated, while an empty string would clear any existing description value. I can't think of any specific problems with _de_-serializing empty as null, but I'm concerned there could be some edge case in the future. Is there a benefit to doing this?

rs/src/contracts/tunnel_environments.rs
```diff
@@ -0,0 +1,39 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+use crate::contracts::tunnel_service_properties::*;
+use url::Url;
+
+pub struct TunnelEnvironment {
```
jasongin, line 7: How is this different from `TunnelServiceProperties`?

rs/src/management/http_client.rs
```diff
@@ -0,0 +1,560 @@
...
+    env_production, Tunnel, TunnelConnectionMode, TunnelEndpoint, TunnelEnvironment, TunnelPort,
+};
+
+use super::{
+    Authorization, HttpError, HttpResult, ResponseError, TunnelLocator, TunnelRequestOptions,
+};
+
+pub struct TunnelManagementClient {
+    client: Client,
+    authorization: Option<HeaderValue>,
```
jasongin, line 21: Other languages have an async callback for getting the auth header, in case a token needs to be refreshed or something like that. I don't know what would be the best way to handle that in Rust.

rs/src/management/http_client.rs
```diff
@@ -0,0 +1,560 @@
...
+            let request_id = res
+                .headers()
+                .get("VsSaaS-Request-Id")
+                .and_then(|h| h.to_str().ok())
+                .map(|s| s.to_owned());
+
+            Err(HttpError::ResponseError(ResponseError {
+                url: url_clone,
+                status_code: res.status(),
+                data: res.text().await.ok(),
```
jasongin, line 299: Most error responses should include a JSON `ProblemDetails` object. Should that be deserialized here?

cs/src/Management/TunnelManagementClient.cs
```diff
@@ -400,10 +400,12 @@ private static void ValidateHttpHandler(HttpMessageHandler httpHandler)
                     case HttpStatusCode.Forbidden:
                         var ex = new UnauthorizedAccessException(errorMessage, hrex);
 
-                        if (response.Headers.WwwAuthenticate?.Count > 0)
+                        // The HttpResponseHeaders.WwwAuthenticate property does not correctly
```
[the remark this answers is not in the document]
connor4312: I merged in main to my branch to avoid conflicts. If main is merged to your dev branch, these commits will go away. But I can retarget this to `main` if you're okay with that for a 'beta' client 🙂

cs/tools/TunnelsSDK.Generator/RustContractWriter.cs
connor4312: I noticed that when creating a tunnel, the `name` and `description` come back as an empty string (rather than null/omitted) if not set.

```json
{
  "clusterId": "usw2",
  "tunnelId": "q4zmg8zl",
  "name": "",
  "description": "",
  "tags": [],
  "accessTokens": {
    "host": "eyJhbGciOiJSUzI1NiIsImtpZCI6IjNCQ0REM0ZEQTlGQzY0ODYzMjkxN0E1QjQxQzFERTJCRjIwQTU5NjUiLCJ4NXQiOiJPODNUX2FuOFpJWXlrWHBiUWNIZUtfSUtXV1UiLCJ0eXAiOiJKV1QifQ.eyJjbHVzdGVySWQiOiJ1c3cyIiwidHVubmVsSWQiOiJxNHptZzh6bCIsInNjcCI6Imhvc3QiLCJleHAiOjE2NTU1ODQwMTAsImlzcyI6Imh0dHBzOi8vdHVubmVscy5hcGkudmlzdWFsc3R1ZGlvLmNvbS8iLCJhdWQiOiJodHRwczovL3R1bm5lbHMuYXBpLnZpc3VhbHN0dWRpby5jb20vIiwianRpIjoiOGI4N2VmYjgtYzJlMi00MGJhLWE5YTYtYzQ2NDhjN2ExMGM1IiwiaWF0IjoxNjU1NDk3NjEwLCJuYmYiOjE2NTU0OTY3MTB9.VvWZe0n3YEYRURZyatTaEpICTaZOtnEYgSMFmHL-nhUvsRx_CSVlJHallyyRBdBIU3_Pyr2v2jwl0CBMwje7p5Ym3puYwtv3EXZzTxBkAoeboIQuDYxeYNFAY6gzTU9HZertiFFQVdyHskaAmwr8DT47gFprHr1-bgcBWu4rMnbjtiBGyc4kOmR04B9i7qioKRdR6NfQKN_L2KmTtK1sC_hFsfgfV5SktFsEJrr8iWbQpp181JE9y7WiIZN0vnQ0HdNyav9Og_8UCe8NAZ_wkJsCiaWrc4u6oBWg0OK7lFIWDqH4nClQZjZcRzag-0RKZKOJY8YFEx-quMS7YzHOPA"
  },
  "accessControl": { "entries": [] },
  "options": {},
  "status": {},
  "endpoints": [],
  "created": "2022-06-17T20:26:50.5466613Z"
}
```

This caused issues when we tried to generate the path for the tunnel, since we'd see there was `Some` name and try to refer to it using that name.

rs/src/contracts/tunnel_environments.rs
connor4312: Ah, missed that. Mostly the same except that it uses a parsed `Url` for the service uri, rather than a String. Could get rid of the duplication.

cs/tools/TunnelsSDK.Generator/RustContractWriter.cs
jasongin: Yes, a tunnel always has a description, and empty string is the default value.

An unnamed tunnel has a name value of empty string rather than null so that the update semantics can work. For example if a tunnel has a (nonempty) name, then you want to update it to remove the name, you'd send an update request with name="", because a request with name=null would indicate no change to the name.

This could be resolved another way by supporting the PATCH verb with partial JSON objects, so that updates would entirely omit properties that aren't meant to be changed. But that's more complicated for all the clients to deal with, so we haven't implemented it.

connor4312: In [JSON Merge Patch](https://datatracker.ietf.org/doc/html/rfc7386), it treated that sending `name: null` would remove the name, and omitting it from the request would leave it unchanged in an update. However not all deserializers have a way to distinguish null/omitted...

But I will remove the empty string==None logic 🙂

jasongin: > However not all deserializers have a way to distinguish null/omitted...

Right, and not all serializers have a way to omit properties. So while it could work, it adds complexity that I think wouldn't be justified at this time.

rs/src/management/http_client.rs
connor4312: Async is rust is always fun, but I implemented that 😬  Consumers can now pass something like I do in tests:

```rs
        c.authorization_fn(|| async {
            let token = match env::var("TUNNEL_TEST_AAD_TOKEN") {
                Ok(value) => value,
                _ => do_device_code_flow(&reqwest::Client::new()).await,
            };

            env::set_var("TUNNEL_TEST_AAD_TOKEN", &token);
            Ok(Authorization::Bearer(token))
        });
```

cs/src/Management/TunnelManagementClient.cs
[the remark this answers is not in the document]
jasongin: Yes I think it's fine for this to merge into `main` even though it is incomplete, since the part that's done is useful. You/we can iterate on it further there.

Please re-target to `main` and then I'll approve the PR.

connor4312:
Looks like the build failure is an issue with auth in the pipeline, not related to my changes

jasongin (approved):

jasongin:
I'm looking into the C# pipeline failure.

jasongin:
It might be a permission issue because the PR submitter is not on the team.

I'm going to override and merge this now. (I can since I'm an admin.) Longer term, we'll eliminate dependencies on internal package feeds, so this won't be an issue after that.

jasongin (approved):

PR #93, opened 2022-06-24
Add service tag access control entry

jessetrinity:
Adds a service tag provider to the IP address range access control entry type.

### DIFF
--- a/cs/src/Contracts/TunnelAccessControlEntryType.cs
+++ b/cs/src/Contracts/TunnelAccessControlEntryType.cs
@@ -54,15 +54,8 @@ namespace Microsoft.VsSaaS.TunnelService.Contracts
 
         /// <summary>
         /// The access control entry is a list of IP address ranges that are allowed (or denied)
-        /// access to the tunnel.
+        /// access to the tunnel. Ranges can be IPv4, IPv6, or Azure service tags.
         /// </summary>
         IPAddressRanges,
-
-        /// <summary>
-        /// The access control entry is a list of service tags representing Azure services that
-        /// are allowed or denied access to the tunnel. These are resolve to IP address ranges by
-        /// the tunnel service.
-        /// </summary>
-        ServiceTags,
     }
 }
--- a/cs/src/Contracts/TunnelAccessControlEntryType.cs
+++ b/cs/src/Contracts/TunnelAccessControlEntryType.cs
@@ -63,6 +63,6 @@ namespace Microsoft.VsSaaS.TunnelService.Contracts
         /// are allowed or denied access to the tunnel. These are resolve to IP address ranges by
         /// the tunnel service.
         /// </summary>
-        ServiceTag,
+        ServiceTags,
     }
 }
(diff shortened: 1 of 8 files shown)


go/tunnels/tunnel_access_control_entry.go
```diff
@@ -75,17 +75,20 @@ type TunnelAccessControlEntryProvider string
 
 const (
 	// Microsoft (AAD) identity provider.
-	TunnelAccessControlEntryProviderMicrosoft TunnelAccessControlEntryProvider = "microsoft"
+	TunnelAccessControlEntryProviderMicrosoft  TunnelAccessControlEntryProvider = "microsoft"
 
 	// GitHub identity provider.
-	TunnelAccessControlEntryProviderGitHub    TunnelAccessControlEntryProvider = "github"
+	TunnelAccessControlEntryProviderGitHub     TunnelAccessControlEntryProvider = "github"
```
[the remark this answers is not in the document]
jasongin: It's just adjusting the whitespace alignment since the new member name was slightly longer than the longest name before.

cs/src/Contracts/TunnelAccessControlEntryType.cs
```diff
@@ -57,5 +57,12 @@ public enum TunnelAccessControlEntryType
         /// access to the tunnel.
         /// </summary>
         IPAddressRanges,
+
+        /// <summary>
+        /// The access control entry is a list of service tags representing Azure services that
+        /// are allowed or denied access to the tunnel. These are resolve to IP address ranges by
+        /// the tunnel service.
+        /// </summary>
+        ServiceTag,
```
jasongin: Would it make sense to _not_ add a new entry type, and use `IPAddressRanges` entry type also for service tags? Then the range format depends on whether the provider name is "ipv4", "ipv6", or "service-tag".

jessetrinity: We can, and I've gone back and forth on it. Having a separate entry just makes the logic a little cleaner when we do the actual access calculation. If one entry seems better to you I'm fine changing it.

derekbekoe: Whichever way we go here, I'd like to support multiple service tags.
Is that already accounted for?

jessetrinity: Yes, the name was meant to be plural here. Should be fixed now.

jasongin: I'd lean toward using the `IPAddressRanges` type, since services tags are just another way to specify IP ranges. Unless you think it will really be a problem.

> Whichever way we go here, I'd like to support multiple service tags. Is that already accounted for?

Yes, access control entry objects already support an array of "subjects" per entry. For `Type=IPAddressRanges` and `Provider=service-tag`, each subject array item would be one service tag.

jessetrinity: Updated to use `IPAddressRanges` for both.

jasongin (approved):

PR #96, opened 2022-06-29
Add name, description, & tags props to TunnelPort

jasongin:
This change adds `name`, `description`, and `tags` properties to the `TunnelPort` contract. These properties will give applications more options for working with multiple ports of a tunnel. They're currently ignored by the tunnel service, but will be implemented later.

### DIFF
--- a/rs/src/contracts/tunnel_access_control_entry.rs
+++ b/rs/src/contracts/tunnel_access_control_entry.rs
@@ -26,8 +26,9 @@ pub struct TunnelAccessControlEntry {
     // the user/group/org IDs. It may be one of the well-known provider names in
     // `TunnelAccessControlEntry.Providers`, or (in the future) a custom identity
     // provider.  For public key ACEs, this value is the type of public key, e.g. "ssh".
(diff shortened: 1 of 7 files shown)


rs/src/contracts/tunnel_access_control_entry.rs
```diff
@@ -26,8 +26,9 @@ pub struct TunnelAccessControlEntry {
     // the user/group/org IDs. It may be one of the well-known provider names in
     // `TunnelAccessControlEntry.Providers`, or (in the future) a custom identity
     // provider.  For public key ACEs, this value is the type of public key, e.g. "ssh".
-    // For IP address range ACEs, this value is the IP addrss version, e.g. "ipv4" or
-    // "ipv6".  For anonymous ACEs, this value is null.
+    // For IP address range ACEs, this value is the IP address version, "ipv4" or "ipv6",
```
jessetrinity, line 29: Looks like this was missed in the related PR - does the rust contract generator not run automatically?

jasongin: It runs during the C# SDK build. It should have been updated since there were C# changes in that PR.

PR #98, opened 2022-07-13
Move search tags to request options

jasongin:
The "search" API was originally designed to search tunnels by tags across all users, returning any matching tunnels that the caller had some access to. But that kind of search is too expensive, and not useful in any known scenarios. This update replaces that function with a more limited option to include tags as a filter to the "list" API that only covers tunnels owned by the caller. And since ports will also soon support tags, and tags are just one of many query options, I'm moving the tags option to the general request options class.

 - Mark the `SearchTunnelsAsync()` API as `[Obsolete]` in C#; remove it in the other 4 languages. We'll remove it from C# in the future.
 - Add `Tags` and `RequireAllTags` to `TunnelRequestOptions`, and update `ToQueryString()` to include those parameters (all 5 languages).
 - Add `ForceRename` request option where it was missing in TS, Java, and Rust.

### DIFF
--- a/cs/src/Management/ITunnelManagementClient.cs
+++ b/cs/src/Management/ITunnelManagementClient.cs
@@ -18,7 +18,7 @@ namespace Microsoft.VsSaaS.TunnelService
     public interface ITunnelManagementClient : IDisposable
     {
         /// <summary>
-        /// Lists all tunnels that are owned by the caller.
+        /// Lists tunnels that are owned by the caller.
         /// </summary>
         /// <param name="clusterId">A tunnel cluster ID, or null to list tunnels globally.</param>
         /// <param name="domain">Tunnel domain, or null for the default domain.</param>
@@ -27,6 +27,11 @@ namespace Microsoft.VsSaaS.TunnelService
         /// <returns>Array of tunnel objects.</returns>
         /// <exception cref="UnauthorizedAccessException">The client access token was missing,
         /// invalid, or unauthorized.</exception>
+        /// <remarks>
+        /// The list can be filtered by setting <see cref="TunnelRequestOptions.Tags"/>.
+        /// Ports will not be included in the returned tunnels unless
+        /// <see cref="TunnelRequestOptions.IncludePorts"/> is set to true.
+        /// </remarks>
         Task<Tunnel[]> ListTunnelsAsync(
             string? clusterId = null,
             string? domain = null,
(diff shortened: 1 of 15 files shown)


jfullerton44 (approved):

connor4312:
Rust changes lgtm

rs/src/management/tunnel_request_options.rs
```diff
@@ -17,19 +17,33 @@ pub struct TunnelRequestOptions {
     /// Gets or sets a flag that requests tunnel ports when retrieving a tunnel object.
     pub include_ports: bool,
 
+    /// Gets or sets an optional list of tags to filter the requested tunnels or ports.
+    pub tags: Vec<String>,
+
+    /// Gets or sets a flag that indicates whether listed items must match all tags
+    /// specified in `tags`. If false, an item is included if any tag matches.
+    pub require_all_tags: bool,
```
connor4312, line 25: As a small tweak I might make this an

```
pub enum SearchTags {
    Any(Vec<String>),
    All(Vec<String>),
}
```

since `require_all_tags` is a no-op without `tags` being specified.

Note this would require a manual implementation of Default, like

```
impl Default for SearchTags {
   pub fn default() -> Self {
        SearchTags::Any(vec![])
    }
}
```

jasongin: Interesting suggestion. I often don't recognize all the applications for Rust's more powerful enums. But in this case it seems like it adds a bit of complexity (additional enum type and necessity of referencing it to specify tags) without bringing much value. And it makes the Rust tunnel API model a little more different from other SDK languages, which (cumulatively) will add to maintenance overhead. So for those reasons I'll decline the suggestion.

cs/src/Management/ITunnelManagementClient.cs
```diff
@@ -27,6 +27,11 @@ public interface ITunnelManagementClient : IDisposable
         /// <returns>Array of tunnel objects.</returns>
         /// <exception cref="UnauthorizedAccessException">The client access token was missing,
         /// invalid, or unauthorized.</exception>
+        /// <remarks>
+        /// The list can be filtered by setting <see cref="TunnelRequestOptions.Tags"/>.
```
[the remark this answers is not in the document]
jasongin: I updated the PR to mention `requireAllTags` in the `tags` doc-comments. (I don't think it's important enough to mention here.)

derekbekoe (approved):

PR #99, opened 2022-07-14
Make SDK cs client reconnect SSH host and client sessions

IlyaBiryukov:
Fix for [#412](https://github.com/microsoft/basis-planning/issues/412) and [#413](https://github.com/microsoft/basis-planning/issues/413)

- Enable SSH session reconnection.
- Retry connection to Relay server when it fails, or is disconnected with exponential backoffs.
- Do not retry on unrecoverable errors.
- Handle 401 errors and fire a new event to let the customers supply a fresh tunnel access token.
- Abstract relay connection logic into a separate class and use it from the tunnel client and host
- Add a common base class for the tunnel client and host, and add an event for tunnel access token refresh there.
- Add some unit tests; more incoming later.

### DIFF
--- a/cs/src/Connections/ConnectionStatusChangedEventArgs.cs
+++ b/cs/src/Connections/ConnectionStatusChangedEventArgs.cs
@@ -21,6 +21,7 @@ public class ConnectionStatusChangedEventArgs : EventArgs
     {
         PreviousStatus = previousStatus;
         Status = status;
+        DisconnectException = disconnectException;
     }
 
     /// <summary>
--- a/cs/src/Connections/ConnectionStatusChangedEventArgs.cs
+++ b/cs/src/Connections/ConnectionStatusChangedEventArgs.cs
@@ -1,0 +1,40 @@
+﻿// <copyright file="ConnectionStatusChangedEventArgs.cs" company="Microsoft">
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license.
+// </copyright>
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.VsSaaS.TunnelService;
+
+/// <summary>
+/// Connection status change event args.
+/// </summary>
+public class ConnectionStatusChangedEventArgs : EventArgs
+{
+    /// <summary>
+    /// Create a new instance of <see cref="ConnectionStatusChangedEventArgs"/>.
+    /// </summary>
+    public ConnectionStatusChangedEventArgs(ConnectionStatus previousStatus, ConnectionStatus status, Exception? disconnectException)
+    {
+        PreviousStatus = previousStatus;
+        Status = status;
+    }
+
+    /// <summary>
+    /// Get the previous connection status.
+    /// </summary>
+    public ConnectionStatus PreviousStatus { get; }
+
+    /// <summary>
+    /// Get the current connection status.
+    /// </summary>
+    public ConnectionStatus Status { get; }
+
+    /// <summary>
+    /// Get the exception that caused disconnect if <see cref="Status"/> is <see cref="ConnectionStatus.Disconnected"/>.
+    /// </summary>
+    public Exception? DisconnectException { get; }
+}
--- a/cs/src/Connections/ITunnelClient.cs
+++ b/cs/src/Connections/ITunnelClient.cs
@@ -110,7 +110,7 @@ namespace Microsoft.VsSaaS.TunnelService
         /// Event handler for refreshing the tunnel access token.
         /// The tunnel client will fire this event when it is not able to use the access token it got from the tunnel.
         /// </summary>
-        EventHandler<RefreshingTunnelAccessTokenEventArgs>? RefreshingTunnelAccessToken { get; set; }
+        event EventHandler<RefreshingTunnelAccessTokenEventArgs>? RefreshingTunnelAccessToken;
 
         /// <summary>
         /// Connection status changed event.
--- a/cs/src/Connections/ITunnelClient.cs
+++ b/cs/src/Connections/ITunnelClient.cs
@@ -36,6 +36,17 @@ namespace Microsoft.VsSaaS.TunnelService
         bool AcceptLocalConnectionsForForwardedPorts { get; set; }
 
         /// <summary>
+        /// Gets the connection status.
+        /// </summary>
+        ConnectionStatus ConnectionStatus { get; }
+
+        /// <summary>
+        /// Gets the exception that caused disconnection.
+        /// Null if not yet connected or disconnection was caused by disposing of this object.
+        /// </summary>
+        Exception? DisconnectException { get; }
+
+        /// <summary>
         /// Connects to a tunnel.
         /// </summary>
         /// <param name="tunnel">Tunnel to connect to.</param>
@@ -94,5 +105,16 @@ namespace Microsoft.VsSaaS.TunnelService
         /// <returns>A <see cref="Task{Stream}"/> representing the result of the asynchronous operation.</returns>
         /// <exception cref="InvalidOperationException">If the tunnel is not yet connected and hasn't started connecting.</exception>
         Task<Stream?> ConnectToForwardedPortAsync(int forwardedPort, CancellationToken cancellation);
+
+        /// <summary>
+        /// Event handler for refreshing the tunnel access token.
+        /// The tunnel client will fire this event when it is not able to use the access token it got from the tunnel.
+        /// </summary>
+        EventHandler<RefreshingTunnelAccessTokenEventArgs>? RefreshingTunnelAccessToken { get; set; }
+
+        /// <summary>
+        /// Connection status changed event.
+        /// </summary>
+        event EventHandler<ConnectionStatusChangedEventArgs>? ConnectionStatusChanged;
     }
 }
--- a/cs/src/Connections/ITunnelHost.cs
+++ b/cs/src/Connections/ITunnelHost.cs
@@ -93,7 +93,7 @@ namespace Microsoft.VsSaaS.TunnelService
         /// Event handler for refreshing the tunnel access token.
         /// The tunnel client will fire this event when it is not able to use the access token it got from the tunnel.
         /// </summary>
-        EventHandler<RefreshingTunnelAccessTokenEventArgs>? RefreshingTunnelAccessToken { get; set; }
+        event EventHandler<RefreshingTunnelAccessTokenEventArgs>? RefreshingTunnelAccessToken;
 
         /// <summary>
         /// Connection status changed event.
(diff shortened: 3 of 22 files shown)


IlyaBiryukov:
There is still some issue that I'm investigating with this change. I'll get back to this PR.

cs/src/Connections/TunnelBase.cs
```diff
@@ -0,0 +1,126 @@
...
+    /// SSH session reconnected event.
+    /// If the session doesn't support reconnects, the client may try to start a brand new SSH session.
+    /// This event is fired after either reconnection succeeds or a new session is created.
+    /// </summary>
+    public event EventHandler? SshSessionReconnected;
+
+    /// <summary>
+    /// Gets the fresh tunnel access token when Relay service returns 401.
+    /// </summary>
+    protected async Task<string?> RefreshTunnelAccessTokenAsync(CancellationToken cancellation)
```
jasongin, line 62: There should be a default behavior that uses the `ITunnelManagementClient`'s user credentials to get a fresh access token, if the management client has a user token callback.

Currently `TunnelHostBase` has a management client instance; `TunnelClientBase` does not but maybe it should.

cs/src/Connections/TunnelBase.cs
```diff
@@ -0,0 +1,126 @@
...
+    /// <summary>
+    /// Get tunnel access scope for this tunnel client or host.
+    /// </summary>
+    protected abstract string TunnelAccessScope { get; }
+
+    /// <summary>
+    /// Event handler for refreshing the tunnel access token.
+    /// The tunnel client or host fires this event when it is not able to use the access token it got from the tunnel.
+    /// </summary>
+    public event EventHandler<RefreshingTunnelAccessTokenEventArgs>? RefreshingTunnelAccessToken;
```
jasongin, line 50: Since the refresh is async, and there would typically be only one refresh handler (or zero), would it be simpler if this was an optional callback instead of an event?

cs/src/Connections/ITunnelClient.cs
```diff
@@ -94,5 +94,18 @@ Task ConnectAsync(Tunnel tunnel, CancellationToken cancellation)
...
+        /// The tunnel client will fire this event when it is not able to use the access token it got from the tunnel.
+        /// </summary>
+        event EventHandler<RefreshingTunnelAccessTokenEventArgs>? RefreshingTunnelAccessToken;
+
+        /// <summary>
+        /// SSH session reconnected event.
+        /// If the session doesn't support reconnects, the client may try to start a brand new SSH session.
+        /// This event is fired after either reconnection succeeds or a new session is created.
+        /// </summary>
+        event EventHandler SshSessionReconnected;
```
jasongin, line 109: Should there be a `Disconnected` event? Or combine them as `ConnectionStatusChanged`?

Also it might be better to drop the `SshSession` prefix, to hide the SSH implementation details as much as possible, in case we ever want to support a new non-SSH connection protocol.

jasongin: I see there is `TunnelClientBase.SshSessionClosed`; maybe that can be merged or at least made consistent.

cs/src/Connections/TunnelHostBase.cs
```diff
@@ -183,22 +192,57 @@ public async Task StartAsync(Tunnel tunnel, CancellationToken cancellation)
 
             // When forwarding from a Remote port we assume that the RemotePortNumber
             // and requested LocalPortNumber are the same.
+            var portNumber = (int)port.PortNumber;
+
+            if (pfs.LocalForwardedPorts.Any((p) => p.LocalPort == portNumber || p.RemotePort == portNumber))
```
jasongin, line 197: I think `p.LocalPort` could be different if the requested port number was in use. Anyway wouldn't it be sufficient to just compare to `p.RemotePort`?

cs/src/Connections/RelayTunnelConnector.cs
```diff
@@ -0,0 +1,221 @@
...
+        {
+            cancellation.ThrowIfCancellationRequested();
+            attempt++;
+            isDelayNeeded = true;
+            Stream? stream = null;
+            errorDescription = null;
+            try
+            {
+                Trace.TraceInformation("Connecting to {0} tunnel relay {1}", Role, this.relayUri.AbsoluteUri);
+
```
jasongin, line 69: Also trace the token properties here, using the `TunelAccessTokenProperties` code I just checked in.

cs/src/Connections/TunnelHostBase.cs
IlyaBiryukov: It needs to check `p.LocalPort` only, not `p.RemotePort`, which may be random if the forwarded port is used on the client machine.

cs/src/Connections/ConnectionStatusChangedEventArgs.cs
```diff
@@ -0,0 +1,40 @@
...
+/// Connection status change event args.
+/// </summary>
+public class ConnectionStatusChangedEventArgs : EventArgs
+{
+    /// <summary>
+    /// Create a new instance of <see cref="ConnectionStatusChangedEventArgs"/>.
+    /// </summary>
+    public ConnectionStatusChangedEventArgs(ConnectionStatus previousStatus, ConnectionStatus status, Exception? disconnectException)
+    {
+        PreviousStatus = previousStatus;
```
jfullerton44, line 22: Is there a reason we dont set disconnectException here?

IlyaBiryukov: No reason. Will fix. Good catch.

cs/src/Connections/RelayTunnelConnector.cs
```diff
@@ -193,6 +168,54 @@ public async Task ConnectSessionAsync(bool isReconnect, CancellationToken cancel
...
+                    }
+                    catch (OperationCanceledException) when (cancellation.IsCancellationRequested)
+                    {
+                        throw;
+                    }
+                    catch (UnauthorizedAccessException uaex)
+                    {
+                        // The refreshed token is not valid.
+                        throw new UnauthorizedAccessException(
+                            $"Not authorized{statusCode}. Refreshed tunnel access token is not valid. {uaex.Message}",
```
jasongin, line 189: Add a space/parentheses: `$"Not authorized ({statusCode})`. Also below.

IlyaBiryukov: It's added on line 174. If the status code is missing, `statusCode` will be and empty string. Otherwise, it is 
`$" ({exception.Data["HttpStatusCode"]})"`.

jasongin: Oh I see, I had missed the extra formatting above.

cs/src/Connections/RefreshingTunnelAccessTokenEventArgs.cs
```diff
@@ -0,0 +1,41 @@
...
+    /// Tunnel access scope to get the token for.
+    /// </summary>
+    public string TunnelAccessScope { get; }
+
+    /// <summary>
+    /// Optional token provider function the event handler may set to asynchnronously fetch the token.
+    /// The arguments are: tunnel access scope and cancellation token.
+    /// The result is the task that returns the new tunnel access token or null if it couldn't fetch the token.
+    /// </summary>
+    public Func<string, CancellationToken, Task<string?>>? TunnelAccessTokenProvider { get; set; }
```
jasongin, line 35: Is there a reason to make this a `Func<..., Task<string>>` instead of a `Task<string?>`?

For events that can return an async result, I have always used the latter. For example in [SshRequestEventArgs](https://devdiv.visualstudio.com/DevDiv/_git/SSH?path=/src/cs/Ssh/Events/SshRequestEventArgs.cs&version=GBmain&line=64&lineEnd=65&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents)

It does require passing the `CancellationToken` as another property on the event.

cs/src/Connections/RefreshingTunnelAccessTokenEventArgs.cs
```diff
@@ -0,0 +1,41 @@
...
+
+    /// <summary>
+    /// Tunnel access token that the event handler may set synchronously.
+    /// </summary>
+    public string? TunnelAccessToken { get; set; }
```
jasongin, line 40: This might be unnecessary if the handler can just provide a synchronous result as `Task.CompletedTask(token)`.

cs/src/Connections/ITunnelHost.cs
```diff
@@ -78,6 +89,15 @@ public interface ITunnelHost : IAsyncDisposable
             TunnelPort updatedPort,
             CancellationToken cancellation);
 
-        // TODO: Events about tunnel port and connection activity?
+        /// <summary>
+        /// Event handler for refreshing the tunnel access token.
+        /// The tunnel client will fire this event when it is not able to use the access token it got from the tunnel.
+        /// </summary>
+        EventHandler<RefreshingTunnelAccessTokenEventArgs>? RefreshingTunnelAccessToken { get; set; }
```
jasongin, line 96: Having an event handler as a get/set property will be confusing. Switching back to an `event` is fine.

cs/src/Connections/RefreshingTunnelAccessTokenEventArgs.cs
jasongin: Oh, this is not what I meant when suggesting a "callback" for the event. For that suggestion there would be a `Func<..., Task<string>>` directly on the `ITunnelHost`. But maybe an event is better.

IlyaBiryukov: I'll change it to `Task<string?>?`

jasongin (approved):

cs/src/Connections/RelayTunnelConnector.cs
```diff
@@ -0,0 +1,222 @@
...
+                        case 403:
+                            throw exception = new UnauthorizedAccessException(
+                                $"Forbidden (403). Provide a fresh tunnel access token with '{this.relayClient.TunnelAccessScope}' scope.", 
+                                wse);
+
+                        case 404:
+                            throw exception = new TunnelConnectionException($"The tunnel or port is not found (404).", wse);
+
+                        case 429:
+                            errorDescription = "Rate limit exceeded (429). Too many requests in a given amount of time.";
```
jasongin, line 115: Probably there should be some delay before retrying after a 429.

I'm currently implementing the server-side code that may return 429. The first limit will be a maximum number of concurrent client connections (not a connection rate). So after a 429 it won't succeed on retry unless some other client has disconnected in the meantime.

jasongin: Oh, I see `isDelayNeeded` will be true at this point, so it's good.

PR #104, opened 2022-07-21
Update SSH package version to get re-forward fix

jasongin:


jfullerton44 (commented):
You need to update ts/src/connections/package.json as well

jfullerton44 (approved):

PR #106, opened 2022-07-27
Add mocha trx reporter

jessetrinity:
This open sources the mocha trx reporter dependency and publishes it to github packages for this repo. See https://github.com/microsoft/dev-tunnels/packages/1563273. 

The ssh package also depends on this package. Once that has been open sourced in this repo as well, the `vs` package feed can also be removed.

### DIFF
--- a/ts-mocha-trx-reporter/.gitignore
+++ b/ts-mocha-trx-reporter/.gitignore
@@ -1,0 +1,3 @@
+node_modules
+dist
+tsconfig.tsbuildinfo
--- a/ts-mocha-trx-reporter/.npmrc
+++ b/ts-mocha-trx-reporter/.npmrc
@@ -1,0 +1,1 @@
+@microsoft:registry=https://npm.pkg.github.com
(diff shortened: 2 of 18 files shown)


jasongin:
This is not a good way to handle a fork of an open-source package... The correct way should be to fork the original repo and then apply the differences. That way the changes can be easily diffed and ported back and forth between the two forks.

jessetrinity:
> This is not a good way to handle a fork of an open-source package... The correct way should be to fork the original repo and then apply the differences. That way the changes can be easily diffed and ported back and forth between the two forks.

If it was indeed a fork of `mocha-trx-reporter` then it was rewritten in TypeScript at some point and bears little resemblance. I don't even see the tests that the original has.

jasongin:
Hmm, I didn't realize it had diverged that much.

Can we consider moving to junit reporting instead, as suggested [here](https://github.com/microsoft/azure-pipelines-tasks/issues/10024#issuecomment-493013509)? Then we won't have to add all this extra code.

PR #107, opened 2022-07-28
Redact tokens from log, check expiration on client

jasongin:
Fixes: https://github.com/microsoft/basis-planning/issues/88
Fixes: https://github.com/microsoft/basis-planning/issues/224#issuecomment-1197070218

 - Add new `TunnelAccessTokenProperties` class that is capable of parsing tunnel access token JWTs. (Only in C# and TS for now.)
 - Check access token expiration date in the client before attempting to use the token in a request.
    - Note this does not apply to AAD/GH user tokens because those are opaque to us: AAD tokens are encrypted and GH tokens are not JWTs.
 - When tracing TS client requests, replace access token with the token properties. The token properties are helpful for diagnostics but don't need to be kept secret like the actual token. Also redact AAD & GH tokens in the Authorization header.
    - C# client tracing is done via an HTTP handler set up by the CLI, so I'll have a following PR in that repo.
 - Rename `accessTokenCallback` to `userTokenCallback` to more clearly distinguish that the purpose of that callback is to obtain AAD/GH tokens rather than tunnel access tokens.

### DIFF
--- a/ts/src/management/tunnelAccessTokenProperties.ts
+++ b/ts/src/management/tunnelAccessTokenProperties.ts
@@ -92,7 +92,6 @@ export class TunnelAccessTokenProperties {
         }
 
         try {
-            const jsonObj = JSON.parse(tokenBodyJson);
             const result = new TunnelAccessTokenProperties();
             Object.assign(result, JSON.parse(tokenBodyJson));
             return result;
--- a/ts/src/management/tunnelAccessTokenProperties.ts
+++ b/ts/src/management/tunnelAccessTokenProperties.ts
@@ -1,0 +1,118 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+/**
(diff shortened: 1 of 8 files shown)


ts/src/management/tunnelAccessTokenProperties.ts
```diff
@@ -0,0 +1,118 @@
...
+            return null;
+        }
+
+        const tokenBodyJson = TunnelAccessTokenProperties.base64UrlDecode(tokenParts[1]);
+        if (!tokenBodyJson) {
+            return null;
+        }
+
+        try {
+            const jsonObj = JSON.parse(tokenBodyJson);
```
IlyaBiryukov, line 95: Unused?

IlyaBiryukov (approved):

ts/src/management/tunnelAccessTokenProperties.ts
jasongin: Removed.

IlyaBiryukov (approved):

PR #108, opened 2022-07-28
rs: fix inherited classes, add tunnel endpoint update method

connor4312:
Noticed when implementing hosting in the Rust SDK that we lacked logic for dealing with inheritance.

### DIFF
--- a/rs/src/management/http_client.rs
+++ b/rs/src/management/http_client.rs
@@ -12,7 +12,7 @@ use url::Url;
 
 use crate::contracts::{
     env_production, Tunnel, TunnelConnectionMode, TunnelEndpoint, TunnelPort,
-    TunnelServiceProperties,
+    TunnelRelayTunnelEndpoint, TunnelServiceProperties,
 };
 
 use super::{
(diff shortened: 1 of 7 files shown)


jasongin (approved):

jasongin (dismissed):

rs/src/management/http_client.rs
```diff
@@ -234,21 +234,22 @@ impl TunnelManagementClient {
...
+    ) -> HttpResult<()> {
         let url = self.build_tunnel_uri(
             locator,
             Some(&format!("{}/{}", PORTS_API_SUB_PATH, port_number)),
         );
         let request = self
             .make_tunnel_request(Method::DELETE, url, options)
             .await?;
-        self.execute_json("create_tunnel_port", request).await
+        self.execute_json::<NeverDeserialize>("delete_tunnel_port", request).await?;
```
jasongin, line 251: The service returns 404 from a request to delete a port that doesn't exist. In the other SDKs that 404 is converted to a `false` return value. Does it make sense to do the same here? Also for `delete_tunnel()`.

rs/src/management/http_client.rs
```diff
@@ -234,21 +234,22 @@ impl TunnelManagementClient {
         self.execute_json("create_tunnel_port", request).await
```
jasongin, line 234: This should be `update` instead of `create`.

rs/src/management/http_client.rs
connor4312: This will return an HTTP error if the service gives back a 404

jasongin (approved):

tanhakabir:
I believe my PR had already incorporated the changes here

PR #109, opened 2022-07-29
Add SshUser to Port and PortSshCommandFormat to TunnelEndpoint

klvnraju:
Required for supporting [CLI changes for SSH Gateway](https://github.com/microsoft/basis-planning/issues/388)
- Port ssh command format to be shown as cli output
- Ssh username for a port to be used at the time of authenticating


cs/src/Contracts/TunnelEndpoint.cs
```diff
@@ -61,6 +61,14 @@ public abstract class TunnelEndpoint
     [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
     public string? PortUriFormat { get; set; }
 
+    /// <summary>
+    /// Gets or sets a string used to format ssh command where ssh client can connect to
+    /// shared ssh port of the tunnel. The string includes a <see cref="PortUriToken"/> that must be
```
jasongin, line 66: Perhaps `PortUriToken` should be renamed since it is not only used for the URI?

jasongin:
There are whitespace differences on every line of many of the files. Is it line endings? And is that intentional? Are some git eol settings wrong for you or for the repo?

klvnraju:
> 

I was not able to figure it out.. trying it on another machine to see if that will make any difference.

cs/src/Contracts/TunnelEndpoint.cs
```diff
@@ -94,6 +102,31 @@ public abstract class TunnelEndpoint
...
+    /// SSH client on Windows/Linux/MacOS  are supported.
+    /// <para />
+    /// If the port is not currently shared via the tunnel, or if a host is not currently
+    /// connected to the tunnel, then ssh connection might fail.
+    /// </remarks>
+    public static string GetPortSshCommand(TunnelEndpoint endpoint, int? portNumber)
+    {
+        if (portNumber == null || string.IsNullOrEmpty(endpoint.PortSshCommandFormat))
+        {
+            return string.Empty;
```
jasongin, line 123: For consistency with `GetPortUri()` maybe this should return `null` in this case. (And change return type to `string?`).

cs/src/Contracts/TunnelPort.cs
```diff
@@ -99,5 +99,14 @@ public TunnelPort()
...
         public TunnelPortStatus? Status { get; set; }
+
+        /// <summary>
+        /// Gets or sets the username for the ssh service user is trying to forward.
+        /// </summary>
+        /// <remarks>
+        /// Should be provided if the <see cref="TunnelProtocol"/> is Ssh.
+        /// </remarks>
+        [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
+        public string? SshUser { get; set; }
```
jasongin, line 110: It's not clear to me how this property is used. Is it set by the user/application when adding the port? In what scenario?

klvnraju: This is used as part of authentication flow from Tunnels Service to the user's ssh server. To give additional background, user authentication from his `ssh client` to `Tunnels Service` happens with his github public ssh key. And after that authentication, `Tunnels Service` has to authenticate with his `sshd` host. And the current way we are supporting this is asking him to add `Tunnels Service` ssh public key to his `~.ssh/authorized_keys` file and provide the corresponding username as this parameter.

IlyaBiryukov (approved):

PR #110, opened 2022-08-05
rs: remove delegated authorization provider

connor4312:
This became very problematic when trying to use the client across
multiple threads; it requires introducing tons of additional constraints
on the futures and I couldn't get it into a place that was workable.

I think it's reasonable for the caller to handle dealing with
authorization. For instance, by having a "bearer" class is called to
return the client and automatically re-acquires credentials at that
time if necessary.

(And having that bearer with delegated authorization present, as implemented,
was not possible since it could not be shared between threads)


jasongin:
The tunnel host and client should support reconnections eventually. For example if the tunnel host loses the connection with the relay, it should try to reconnect, and that may require a refreshed access token. How will that work? I don't think creating a new host/client instance is a good solution because it would lose some state.

(I realize that safety constraints in Rust often force a different object model compared to traditional OO programming, so it can be OK if the approach here differs significantly from the other SDK languages.)

jasongin (approved):

connor4312:
I think that will be workable, for example by implementing an async trait handler like [thrussh](https://docs.rs/thrussh/0.33.5/thrussh/) does for their client. Or just saying the connection is owned by a single thread and not shared.

PR #115, opened 2022-08-12
Report connection error status code

jasongin:
When a websocket connection to the relay fails, check the HTTP status code and use that to generate a better error message in the `TunnelConnectionException`. That exception message will also be displayed by the CLI.

### DIFF
--- a/cs/src/Connections/WebSocketStream.cs
+++ b/cs/src/Connections/WebSocketStream.cs
@@ -6,457 +6,449 @@
 using System;
 using System.Diagnostics;
 using System.IO;
+using System.Net;
 using System.Net.WebSockets;
 using System.Threading;
 using System.Threading.Tasks;
 
-namespace Microsoft.VsSaaS.TunnelService
+namespace Microsoft.VsSaaS.TunnelService;
+
+/// <summary>
+/// Wraps a <see cref="WebSocket"/> to adapt it to <see cref="Stream"/>.
+/// </summary>
+/// <remarks>
+/// Cancelling reads or writes will abort the web socket connection.
+/// If backs SshSession when it is closed, SshSession cancels the read from the stream, which will abort the web socket.
+/// </remarks>
+public class WebSocketStream : Stream
 {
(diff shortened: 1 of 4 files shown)


jasongin:
This needs a rebase...

cs/src/Connections/WebSocketStream.cs
```diff
@@ -6,519 +6,547 @@
...
             {
-                configure?.Invoke(socket.Options);
-                await socket.ConnectAsync(uri, cancellation);
-                return new WebSocketStream(socket);
-            }
-            catch
-            {
-                socket.Dispose();
-                throw;
+                var offset = messageIndex + statusCodeMessage.Length;
```
IlyaBiryukov, line 110: There is already code in `WebSocketStream` that gets the status code from the exception message, checked in with my PR https://github.com/microsoft/dev-tunnels/pull/99

jasongin: Yes, I just realized this is mostly duplicating that. I'll close this PR.

cs/src/Connections/WebSocketStream.cs
```diff
@@ -6,519 +6,547 @@
 using System;
 using System.Diagnostics;
 using System.IO;
+using System.Net;
 using System.Net.WebSockets;
 using System.Threading;
 using System.Threading.Tasks;
 
-namespace Microsoft.VsSaaS.TunnelService
+namespace Microsoft.VsSaaS.TunnelService;
```
IlyaBiryukov, line 14: Lots of changes in this file making it hard to review. Perhaps changing to file-scoped namespace may happen in a separate PR to simplify reviewing this one?

jasongin: For future reference, GitHub can hide whitespace differences:
image

But I closed this PR anyway.

IlyaBiryukov: Thanks for the tip on the whitespace diffing!

PR #116, opened 2022-08-15
Do not check host RemoteForwarders when PFS channel opens

IlyaBiryukov:
There is a small time interval in `ForwardPortAsync()` between the port being forwarded with `ForwardFromRemotePortAsync()` and `RemoteForwarders` updated.

See https://github.com/microsoft/basis-planning/issues/467

The check is redundant because setting `AcceptRemoteConnectionsForNonForwardedPorts` to false makes PFS reject forwarding requests from the clients for the ports that are not forwarded and are missing in `PFS.remoteConnectors`.
Call to `PFS.ForwardFromRemotePortAsync()` in `ForwardPortAsync()` adds the connector to `PFS.remoteConnectors`.

### DIFF
--- a/cs/src/Connections/TunnelRelayTunnelHost.cs
+++ b/cs/src/Connections/TunnelRelayTunnelHost.cs
@@ -458,12 +458,14 @@ namespace Microsoft.VsSaaS.TunnelService
                         return;
                     }
 
-                    if (!RemoteForwarders.ContainsKey(new SessionPortKey(sessionId, (ushort)portForwardRequest.Port)))
-                    {
-                        Trace.Warning("Rejecting request to connect to non-forwarded port:" +
-                            portForwardRequest.Port);
-                        e.FailureReason = SshChannelOpenFailureReason.AdministrativelyProhibited;
-                    }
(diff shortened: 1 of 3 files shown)


cs/src/Connections/TunnelRelayTunnelHost.cs
```diff
@@ -458,12 +458,14 @@ private void OnSshChannelOpening(object? sender, SshChannelOpeningEventArgs e)
                         return;
                     }
 
-                    if (!RemoteForwarders.ContainsKey(new SessionPortKey(sessionId, (ushort)portForwardRequest.Port)))
```
jasongin, line 461: I think the other checks above are unnecessary also, since their primary purpose was to get the `sessionId` for this lookup.

IlyaBiryukov: Done

jasongin (dismissed):

IlyaBiryukov:
@jasongin could you please approve one more time? One PR check was failing because there was a comment with a reference to the bug in the planning repo. I've removed it.

jasongin (approved):

PR #119, opened 2022-08-18
Make TunnelManagementClient extensible

jasongin:
- Define protected `SendTunnelRequestAsync()` and `SendRequestAsync()` methods in C# `TunnelManagementClient` enabling subclasses to support additional management APIs without duplicating a lot of code to format requests and handle responses and errors. (We'll use this for internal APIs.)
 - Apply equivalent changes to TS. (No specific need for this yet but it's good to keep parity with C#.)
 - Fix handling of 404 responses in TS which was not correct for some APIs.

I tried to make the new APIs cleaner and more consistent, and they are well-documented. The PR diff has a lot of changes but much of it is just renaming/reordering parameters, or doc-comments.

### DIFF
--- a/cs/src/Management/TunnelManagementClient.cs
+++ b/cs/src/Management/TunnelManagementClient.cs
@@ -215,41 +215,206 @@ namespace Microsoft.VsSaaS.TunnelService
         private ProductInfoHeaderValue[] UserAgents { get; }
 
         /// <summary>
-        /// Sends an HTTP request for a tunnel, with authorization header from either tunnel
-        /// properties or options.
+        /// Sends an HTTP request to the tunnel management API, targeting a specific tunnel.
         /// </summary>
-        private Task<TResult?> SendTunnelRequestAsync<TResult>(
-            Tunnel? tunnel,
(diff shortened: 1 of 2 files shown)


cs/src/Management/TunnelManagementClient.cs
```diff
@@ -533,45 +708,52 @@ public void Dispose()
...
                 tunnelPath + (!string.IsNullOrEmpty(path) ? path : string.Empty),
-                options,
-                query);
+                query,
+                options);
         }
 
-        private async Task<AuthenticationHeaderValue?> GetAccessTokenAsync(
+        /// <summary>
+        ///
```
IlyaBiryukov, line 716: Empty doc comments.

jasongin: Oh, this one should be `private`. I had considered making it `protected` but there's no need. I'll change back to `private` and remove the empty doc-comments.

IlyaBiryukov (approved):

jessetrinity (approved):

PR #120, opened 2022-08-23
API to make host refresh ports from service

jasongin:
Fixes: https://github.com/microsoft/basis-planning/issues/450

These changes enable a client with management access to update the set of forwarded ports on an actively hosted tunnel.

 - Add `RefreshPortsAsync()` API to `ITunnelHost` and `ITunnelClient`
 - Implement on the host side by using the management client to get the updated list of ports, and updating SSH port-forwarding state accordingly.
 - Implement on the client side by sending a special SSH request message of type `RefreshPorts` to the host, that simply triggers the above operation.
 - Remove Add/Update/Remove ports APIs from `ITunnelHost`. Any calls to those should be replaced by calls to the equivalent management APIs, followed by `RefreshPortsAsync()`. Test code is updated accordingly.

 **TODO**: (soon) Port changes to other SDK languages (Java, Go, Rust). Merging the C# + TS changes first to unblock partners.

 **TODO**: (later) It would be better if the tunnel service could notify the host immediately whenever changes happen. That would eliminate the need for this "refresh" API. But it would require a lot more server-side plumbing to accomplish.

### DIFF
--- a/cs/src/Connections/TunnelHost.cs
+++ b/cs/src/Connections/TunnelHost.cs
@@ -13,220 +13,183 @@ using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.VisualStudio.Ssh;
 using Microsoft.VisualStudio.Ssh.Algorithms;
+using Microsoft.VisualStudio.Ssh.Messages;
 using Microsoft.VisualStudio.Ssh.Tcp;
 using Microsoft.VsSaaS.TunnelService.Contracts;
 
-namespace Microsoft.VsSaaS.TunnelService
+namespace Microsoft.VsSaaS.TunnelService;
+
+/// <summary>
+/// Base class for Hosts that host one tunnel
+/// </summary>
+public abstract class TunnelHost : TunnelConnection, ITunnelHost
 {
+    internal const string RefreshPortsRequestType = "RefreshPorts";
+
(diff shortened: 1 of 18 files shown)


cs/src/Connections/TunnelHost.cs
```diff
@@ -13,220 +13,183 @@
...
             {
-                // The port is already forwarded. This may happen if we try to add the same port twice after reconnection.
-                return;
+                var key = new SessionPortKey(session.SessionId!, port.PortNumber);
+                if (!RemoteForwarders.ContainsKey(key))
+                {
+                    // Overlapping refresh operations could cause duplicate forward requests to be
+                    // sent to clients, but clients should ignore the duplicate requests.
+                    var pfs = session.GetService<PortForwardingService>() !;
+                    await ForwardPortAsync(pfs, port, cancellation);
```
jasongin, line 151: Calls to `ForwardPortAsync()` should be done concurrently. I'll update this code to add tasks to a list and then wait on all of them at once.

jasongin: Done.

IlyaBiryukov (approved):

IlyaBiryukov (approved):

PR #121, opened 2022-08-23
TS tunnel SDK SSH reconnection

IlyaBiryukov:
Add SSH reconnection to ts tunnel SDK.

Refactor ts tunnel SDK:
Extract connection flow to relayTunnelConnector. It implements TunnelConnector and uses TunnelSession interface to abstract tunnel client and host.
Extract common host and client interface into TunnelConnection.
Extract common client and host code into TunnelConnectionBase class.
TunnelConnectionSession extends TunnelConnectionBase and drives connecting to the tunnel.
Extract relay connectivity into mixin class TunnelRelaySessionClass.
Extract sshSession property and disposal into mixin class TunnelSshSessionClass.
Get status code from failed web socket connection.
Bump vs-ssh dependency to 3.7.6.

Add more unit tests and fix test failures.


IlyaBiryukov:
Reconnection unit tests are coming soom.

IlyaBiryukov:
Still TODO: update the `@vs/vs-ssh` and `@vs/vs-ssh-tcp` npm packages version once the new version has been published.

ts/src/connections/utils.ts
```diff
@@ -19,3 +22,92 @@ export class List {
...
+}
+
+/**
+ * Checks if {@link e} is {@link CancellationError} and {@link cancellation} has been requested.
+ */
+export function isCancellation(e: any, cancellation?: CancellationToken): e is CancellationError {
+    return (
+        e &&
+        typeof e === 'object' &&
+        e instanceof CancellationError &&
```
jasongin, line 64: nit: `instanceof` makes both of the prior checks unnecessary. A null/undefined or non-object value will always fail an `instanceof` check.

ts/src/connections/utils.ts
```diff
@@ -19,3 +22,92 @@ export class List {
...
+        (!cancellation || cancellation.isCancellationRequested)
+    );
+}
+
+/**
+ * Checks if {@link e} is {@link Error}.
+ */
+export function isError(e: any): e is Error {
+    return e && typeof e === 'object' && e instanceof Error;
```
jasongin, line 73: `e instanceof Error` is sufficient, making this `isError()` function not much more concise.

ts/src/connections/utils.ts
```diff
@@ -19,3 +22,92 @@ export class List {
...
+ */
+export function getErrorMessage(e: any) {
+    return String(e?.message ?? e);
+}
+
+/**
+ * Wraps e in Error object if e is not Error. If e is Error, returns e as is.
+ */
+export function getError(e: any, messagePrefix?: string): Error {
+    return isError(e) ? e : new Error(`${messagePrefix ?? ''}${getErrorMessage(e)}`);
```
jasongin, line 87: At this point `e` is known to _not_ be an `Error` object -- so why check for a `.message` property?

ts/src/connections/index.ts
```diff
@@ -18,3 +18,11 @@ export * from './tunnelRelayStreamFactory';
 export * from './defaultTunnelRelayStreamFactory';
 export * from './tunnelRelayTunnelClient';
 export * from './tunnelRelayTunnelHost';
+export * from './tunnelConnection';
+export * from './tunnelConnectionBase';
+export * from './tunnelSession';
+export * from './connectionStatus';
```
jasongin, line 24: Does the `TunnelSession` class need to be public?

jasongin: It has methods on it that look like they should be internal implementation details, not callable by applications.

ts/src/connections/tunnelRelayTunnelClient.ts
```diff
@@ -1,87 +1,73 @@
...
     }
 
-    public async connectClient(tunnel: Tunnel, endpoints: TunnelEndpoint[]): Promise<void> {
-        let tunnelEndpoints = endpoints.map((endpoint) => endpoint as TunnelRelayTunnelEndpoint);
+    /**
+     * Gets the tunnel relay URI.
+     */
+    public async getTunnelRelayUri(tunnel?: Tunnel): Promise<string> {
+        let tunnelEndpoints = this.endpoints!.map(
+            (endpoint) => endpoint as TunnelRelayTunnelEndpoint,
```
jasongin, line 31: This `map()` call will return `null` items for enpoints that are not the expected type. But it is intended to filter them out instead. So, this should be changed to `this.endpoints!.filter((endpoint) => endpoint instanceof TunnelRelayTunnelEndpoint)`

(I see that this bug was already there before the PR.)

ts/src/connections/tunnelRelayTunnelClient.ts
```diff
@@ -1,87 +1,73 @@
...
-            throw new Error('Failed to connect to tunnel relay. ' + ex);
+        this.relayUri = clientRelayUri;
+        this.accessToken = accessToken;
+        await this.connectTunnelSession();
+    }
+
+    /**
+     * Configures the tunnel session with the given stream.
+     */
+    public async configureSession(
```
jasongin, line 62: Should this be internal? In TS, internal methods on exported classes can be marked with `/* @internal */` to prevent them from being included in the `.d.ts` type definitions. See https://www.typescriptlang.org/tsconfig#stripInternal

Maybe some other methods should be marked that way also.

ts/src/connections/retryingTunnelConnectionEventArgs.ts
```diff
@@ -0,0 +1,25 @@
...
+
+/**
+ * Event args for tunnel connection retry event.
+ */
+export class RetryingTunnelConnectionEventArgs {
+    constructor(
+        /**
+         * Gets the error that caused the retry.
+         */
+        public readonly error: any,
```
jasongin, line 12: Can it be `Error` type? If somehow we catch a non-Error error, we should probably wrap it in an `Error`.

ts/src/connections/sshHelpers.ts
```diff
@@ -2,12 +2,33 @@
...
+         * 'httpResponse' event in WebSocketClient is fired when the server responds but the HTTP request doesn't properly upgrade to a web socket,
+         * i.e. the status code is not 101 `Switching Protocols`. The argument of the event callback is the recieved response.
+         */
+        on(event: 'httpResponse', cb: (response: IncomingMessage) => void): this;
+    }
+}
+
+/**
+ * Error class for errors connecting to a web socket in non-node (browser) context.
+ * There is no status code or underlying network error info in the browser context.
```
jasongin, line 25: I wonder if the relay service should do something different for rejecting browser websocket connection attempts, in order to propagate error info. Maybe it could accept the websocket and then immediately close it with a websocket status code.

Not blocking this PR, but maybe worth considering later.

jasongin: I see the link below to https://websockets.spec.whatwg.org/#eventdef-websocket-error gives some reasoning for not supplying any error details, but I'm not convinced that reasoning applies to our scenario.

ts/src/connections/sshHelpers.ts
```diff
@@ -136,43 +161,37 @@ export class SshHelpers {
             client.on('connect', (connection: any) => {
                 resolve(new ssh.WebSocketStream(new WebsocketStreamAdapter(connection)));
             });
-            client.on('connectFailed', (e: any) => {
-                if (e.message && e.message.startsWith('Error: ')) {
-                    e.message = e.message.substr(7);
+
+            // If the server responds but doesn't properly upgrade the connection to web socket, WebSocketClient fires 'httpResponse' event.
+            // TODO: Return ProblemDetails from TunnelRelay service
+            client.on('httpResponse', ({ statusCode, statusMessage }) => {
```
jasongin, line 167: Will calling `on('httpResponse')` throw an error in a browser environment where that event is not supported?

ts/src/connections/sshHelpers.ts
```diff
@@ -285,8 +306,74 @@ export class RelayConnectionError extends Error {
...
+        statusCode: 404,
+        error: 'relayEndpointNotFound',
+        errorType: RelayErrorType.EndpointNotFound,
+    },
+    {
+        regex: /status: 404 There are no listeners connected/,
+        statusCode: 404,
+        error: 'relayListenerOffline',
+        errorType: RelayErrorType.ListenerOffline,
+    },
```
jasongin, line 366: These error message patterns are leftover from old Azure Relay error-handling. I don't think they are valid any more for tunnel relay connections.

jasongin: And I see that's mentioned in the comments above. But it can be deleted now.

ts/src/connections/tunnelConnection.ts
```diff
@@ -0,0 +1,46 @@
...
+import { Event } from 'vscode-jsonrpc';
+import { ConnectionStatus } from './connectionStatus';
+import { ConnectionStatusChangedEventArgs } from './connectionStatusChangedEventArgs';
+import { RefreshingTunnelAccessTokenEventArgs } from './refreshingTunnelAccessTokenEventArgs';
+import { RetryingTunnelConnectionEventArgs } from './retryingTunnelConnectionEventArgs';
+
+/**
+ * Tunnel connection.
+ */
+export interface TunnelConnection {
```
jasongin, line 13: `extends Disposable` ?

ts/src/connections/sshHelpers.ts
IlyaBiryukov: The error may happen upstream to our service, and it won't have a chance to respond.
That's really up to the user agent (browser) as to what data is provided in the web socket error, not sure if we can control that from our SDK.
Anyway, `RelayTunnelConnector` treats this error differently and retries several times before giving up.

ts/src/connections/sshHelpers.ts
IlyaBiryukov: In browser environment a different code branch uses a js web socket, not this custom node module.

ts/src/connections/tunnelConnection.ts
IlyaBiryukov: `Disposable` has `dispose(): void` while here it's `dispose(): Promise<void>`.

ts/src/connections/sshHelpers.ts
jasongin: > The error may happen upstream to our service

In some cases, yes, but for some errors (such as tunnel not found or host not connected) the relay service could return an error code in the websocket close message (which is exposed by browsers: https://developer.mozilla.org/en-US/docs/Web/API/CloseEvent/code) or could even send details over the websocket stream.

ts/src/connections/tunnelConnection.ts
jasongin: Is async disposal really needed? `TunnelConnectionBase.dispose()` currently just returns `Promise.resolve()`.

ts/src/connections/tunnelRelayTunnelHost.ts
```diff
@@ -224,19 +218,42 @@ export class TunnelRelayTunnelHost extends TunnelHostBase {
...
+        ports.forEach(async (port) => {
+            try {
+                await this.forwardPort(pfs, port);
+            } catch (ex) {
+                this.traceError(`Error forwarding port ${port.portNumber}: ${ex}`);
+            }
+        });
+
+        // If a tunnel client reconnects, its SSH session Port Forwarding service may
+        // have remote port forwarders for the ports no longer forwarded.
```
jasongin, line 243: Is this necessary? If host ports were refreshed, any remote forwarders for removed ports would have been dropped and disposed, and that would send a message to the client to stop forwarding the port. The message would be queued while the client is in a disconnected state, and sent after reconnection.

IlyaBiryukov: If SSH session reconnection fails and after that regular connection succeeds, the queued messages won't be resent.

ts/src/connections/tunnelConnection.ts
IlyaBiryukov: Child classes use it. E.g. `TunnelRelayTunnelHost.dipsoe()` awaits termination of all client sessions and deletion of the tunnel endpoint.

ts/src/connections/tunnelRelayTunnelHost.ts
jasongin: After a non-SSH reconnection, any client sessions will be all new and will not have any port-forwarding state from before. None of the existing remote forwarder instances here will match the new session.

ts/src/connections/tunnelRelayTunnelClient.ts
IlyaBiryukov: `instanceof` cannot be used because `TunnelRelayTunnelEndpoint` is an interface. Instead, we can use `endpoint.connectionMode === TunnelConnectionMode.TunnelRelay` condition.

jasongin (approved):

PR #122, opened 2022-08-24
Add cluster details for list clusters API

jessetrinity:
Adds a `ClusterDetails` type and a method to get the list of clusters from the service.


IlyaBiryukov (dismissed):

cs/src/Contracts/ClusterDetails.cs
```diff
@@ -0,0 +1,15 @@
+namespace Microsoft.VsSaaS.TunnelService.Contracts;
+
+/// <summary>
+/// Tunnel service cluster details.
+/// </summary>
+public class ClusterDetails {
```
jasongin, line 6: I'd suggest an `AzureLocation` property that has the standard Azure location name.

cs/src/Contracts/ClusterDetails.cs
```diff
@@ -0,0 +1,15 @@
...
+    /// <summary>
+    /// A cluster identifier based on its region.
+    /// </summary>
+    public string? ClusterId { get; set; }
+    /// <summary>
+    /// The cluster DNS host.
+    /// </summary>
+    public string? Host { get; set; }
```
jasongin, line 14: How about `Hostname` so it is less easily confused with a tunnel "host"?

cs/src/Contracts/ClusterDetails.cs
```diff
@@ -0,0 +1,15 @@
+namespace Microsoft.VsSaaS.TunnelService.Contracts;
+
+/// <summary>
+/// Tunnel service cluster details.
+/// </summary>
+public class ClusterDetails {
+    /// <summary>
+    /// A cluster identifier based on its region.
+    /// </summary>
+    public string? ClusterId { get; set; }
```
jasongin, line 10: Why is it nullable? It should always have a value.

(In some other contracts, some required properties are nullable to support partial updates. But there will not be an API for updating clusters.)

cs/src/Contracts/ClusterDetails.cs
```diff
@@ -0,0 +1,15 @@
+namespace Microsoft.VsSaaS.TunnelService.Contracts;
+
+/// <summary>
+/// Tunnel service cluster details.
```
jasongin, line 4: Include a brief explanation of what a "cluster" is.

cs/src/Contracts/ClusterDetails.cs
jasongin: Or should it be a full `https://` URI?

jessetrinity: I had it with https before which was nicer in the CLI, but I thought host name might be better for URI builders. But URI builders are probably happy taking either form anyway.

jfullerton44 (dismissed):
Go code looks good

cs/src/Contracts/ClusterDetails.cs
jessetrinity: Updated to use full uri.

IlyaBiryukov (approved):

jasongin (approved):

PR #123, opened 2022-08-25
 Add manage:ports access scope

jasongin:
Add SDK support for a new `manage:ports` scope, and for requesting access tokens with a space-delimited combination of scopes, such as `connect manage:ports`.

Note attempts to use the new scope will be rejected by the tunnel service as invalid until the server-side support lands. I'm working on that separately.

Related: https://github.com/microsoft/basis-planning/issues/450

### DIFF
--- a/cs/src/Contracts/TunnelAccessControl.cs
+++ b/cs/src/Contracts/TunnelAccessControl.cs
@@ -78,7 +78,8 @@ namespace Microsoft.VsSaaS.TunnelService.Contracts
         /// <param name="validScopes">Optional subset of scopes to be considered valid;
         /// if omitted then all defined scopes are valid.</param>
         /// <param name="allowMultiple">Whether to allow multiple space-delimited scopes in a
-        /// single item.</param>
+        /// single item. Multiple scopes are supported when requesting a tunnel access token
+        /// with a combination of scopes.</param>
         /// <exception cref="ArgumentException">A scope is not valid.</exception>
         public static void ValidateScopes(
             IEnumerable<string> scopes,
--- a/cs/src/Contracts/TunnelAccessControl.cs
+++ b/cs/src/Contracts/TunnelAccessControl.cs
@@ -77,14 +77,22 @@ namespace Microsoft.VsSaaS.TunnelService.Contracts
         /// <param name="scopes">List of scopes to validate.</param>
         /// <param name="validScopes">Optional subset of scopes to be considered valid;
         /// if omitted then all defined scopes are valid.</param>
+        /// <param name="allowMultiple">Whether to allow multiple space-delimited scopes in a
+        /// single item.</param>
         /// <exception cref="ArgumentException">A scope is not valid.</exception>
         public static void ValidateScopes(
             IEnumerable<string> scopes,
-            IEnumerable<string>? validScopes = null)
+            IEnumerable<string>? validScopes = null,
+            bool allowMultiple = false)
         {
             if (scopes == null)
             {
                 throw new ArgumentNullException(nameof(scopes));
+            }
+
+            if (allowMultiple)
+            {
+                scopes = scopes.SelectMany((s) => s.Split(' '));
             }
 
             foreach (var scope in scopes)
--- a/cs/src/Management/TunnelRequestOptions.cs
+++ b/cs/src/Management/TunnelRequestOptions.cs
@@ -81,21 +81,18 @@ namespace Microsoft.VsSaaS.TunnelService
         public bool RequireAllTags { get; set; }
 
         /// <summary>
-        /// Gets or sets an optional list of scopes that should be authorized when retrieving a
-        /// tunnel or tunnel port object.
-        /// </summary>
-        /// <remarks>
-        /// For service-to-service calls using forwarded client credentials, the scope(s) should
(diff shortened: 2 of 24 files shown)


cs/src/Management/TunnelRequestOptions.cs
```diff
@@ -80,22 +80,19 @@ public class TunnelRequestOptions
...
-        /// Gets or sets an optional list of scopes that should be authorized when retrieving a
-        /// tunnel or tunnel port object.
-        /// </summary>
-        /// <remarks>
-        /// For service-to-service calls using forwarded client credentials, the scope(s) should
-        /// match the action the client is requesting, such as "host" or "connect". This enables
-        /// the calling service to ensure the client is specifically authorized for the supplied
-        /// scope(s), rather than just any scopes permitted by the API.
-        /// </remarks>
-        public string[]? Scopes { get; set; }
```
jasongin, line 93: The `scopes` property is removed here (and in other languages) because it was always an internal-only parameter and now is no longer needed in the SDK since the relay service uses a different internal API for tunnel connections.

ts/src/management/tunnelManagementHttpClient.ts
```diff
@@ -612,8 +622,22 @@ export class TunnelManagementHttpClient implements TunnelManagementClient {
         }
 
         if (!(tunnelAuthentication in headers) && tunnel?.accessTokens) {
-            for (let scope of scopes ?? []) {
-                const accessToken = tunnel.accessTokens[scope];
+            for (let scope of accessTokenScopes ?? []) {
+                let accessToken: string | null = null;
+                for (let scopeAndToken of Object.entries(tunnel.accessTokens)) {
```
IlyaBiryukov, line 627: nit: can be deconstructed inline into `[scope, token]`

IlyaBiryukov (dismissed):

go/tunnels/manager.go
```diff
@@ -56,8 +56,9 @@ var (
 var (
 	manageAccessTokenScope       = []TunnelAccessScope{TunnelAccessScopeManage}
 	hostAccessTokenScope         = []TunnelAccessScope{TunnelAccessScopeHost}
-	hostOrManageAccessTokenScope = []TunnelAccessScope{TunnelAccessScopeManage, TunnelAccessScopeHost}
-	readAccessTokenScope         = []TunnelAccessScope{TunnelAccessScopeManage, TunnelAccessScopeHost, TunnelAccessScopeConnect}
+	managePortsAccessTokenScopes = []TunnelAccessScope{TunnelAccessScopeManage, TunnelAccessScopeManagePorts, TunnelAccessScopeHost}
+	readAccessTokenScopes        = []TunnelAccessScope{TunnelAccessScopeManage, TunnelAccessScopeHost, TunnelAccessScopeConnect}
+	readPortsAccessTokenScopes   = []TunnelAccessScope{TunnelAccessScopeManage, TunnelAccessScopeManagePorts, TunnelAccessScopeHost, TunnelAccessScopeConnect}
```
jessetrinity, line 61: Not sure if the indentation is off on this line or if the diff is just displaying it that way.

go/tunnels/manager_test.go
```diff
@@ -16,19 +16,20 @@ import (
 )
 
 var (
-	serviceUrl           = ServiceProperties.ServiceURI
+	serviceUrl           = "https://localhost:9901/"
```
jessetrinity, line 19: Was this changed for testing?

cs/src/Contracts/TunnelAccessControl.cs
```diff
@@ -77,16 +77,24 @@ public TunnelAccessControl(IEnumerable<TunnelAccessControlEntry> entries)
         /// <param name="scopes">List of scopes to validate.</param>
         /// <param name="validScopes">Optional subset of scopes to be considered valid;
         /// if omitted then all defined scopes are valid.</param>
+        /// <param name="allowMultiple">Whether to allow multiple space-delimited scopes in a
```
jessetrinity, line 80: Should this comment indicate when/why this would be beneficial?

go/tunnels/manager_test.go
jasongin: Oh, yes that change was not supposed to be in the PR.

go/tunnels/manager.go
jasongin: It looks correct to me. Try toggling the ignore-whitespace option on the diff view.

jessetrinity: Looks like it was actually the "split view" that was throwing me off :)

ts/src/management/tunnelManagementHttpClient.ts
jasongin: I considered that, but there are so many scope and token variables already in this part of code that I thought it would be clearer to keep the key-value pair together.

go/tunnels/manager_test.go
jasongin: Reverted.

cs/src/Contracts/TunnelAccessControl.cs
jasongin: Added.

jessetrinity (approved):

PR #127, opened 2022-08-31
rs: implement initial host relay 

connor4312:
This implement a host relay. It does not yet implement reconnections,
although with that added in consumer code this has been working fairly
well for me, and I've been using this for my day-to-date work.
The relay is opt-in under the `connections` feature flag.

This PR is marked as draft until russh publishes their next version,
at which time I'll update the Cargo.toml and this'll be ready to go.

### DIFF
--- a/rs/src/connections/host_relay.rs
+++ b/rs/src/connections/host_relay.rs
@@ -34,6 +34,7 @@ pub struct HostRelay {
     ports_tx: watch::Sender<PortMap>,
     ports_rx: watch::Receiver<PortMap>,
     mgmt: TunnelManagementClient,
+    host_keypair: russh_keys::key::KeyPair,
 }
 
 /// Hello friend. You're probably here because you want to change how tunnel
@@ -127,6 +128,11 @@ impl HostRelay {
             ports_tx,
             ports_rx,
             mgmt,
+            host_keypair: russh_keys::key::KeyPair::generate_rsa(
+                2048,
+                russh_keys::key::SignatureHash::SHA2_512,
+            )
+            .expect("expected to generate rsa keypair"),
         }
     }
 
@@ -166,10 +172,13 @@ impl HostRelay {
         let client_session = Arc::new(client_session);
         let client_session_ret = client_session.clone();
 
+        log::debug!("established host relay primary session");
+
         let mut channels = HashMap::new();
         let ports_rx = self.ports_rx.clone();
+        let host_keypair = self.host_keypair.clone();
         let join = tokio::spawn(async move {
-            let mut server = HostRelay::make_ssh_server();
+            let mut server = HostRelay::make_ssh_server(host_keypair.clone());
             loop {
                 tokio::select! {
                     Some(op) = rx.recv() => match op {
@@ -199,6 +208,8 @@ impl HostRelay {
                 .disconnect(russh::Disconnect::ByApplication, "going away", "en")
                 .await
                 .ok();
+
+            log::debug!("disconnected primary session after EOF");
 
             Ok(())
         });
@@ -301,15 +312,18 @@ impl HostRelay {
         Ok(())
     }
 
-    fn make_ssh_server() -> Server {
+    fn make_ssh_server(keypair: russh_keys::key::KeyPair) -> Server {
         let c = russh::server::Config {
             connection_timeout: None,
             auth_rejection_time: std::time::Duration::from_secs(5),
-            keys: vec![russh_keys::key::KeyPair::generate_rsa(
-                2048,
-                russh_keys::key::SignatureHash::SHA2_512,
-            )
-            .expect("expected to generate rsa keypair")],
+            keys: vec![keypair],
+            window_size: 1024 * 1024 * 64,
+            preferred: russh::Preferred::COMPRESSED,
+            limits: russh::Limits {
+                rekey_read_limit: usize::MAX,
+                rekey_time_limit: Duration::MAX,
+                rekey_write_limit: usize::MAX,
+            },
             ..Default::default()
         };
 
@@ -328,12 +342,13 @@ impl HostRelay {
     > {
         let config = russh::client::Config {
             anonymous: true,
+            window_size: 1024 * 1024 * 64,
             preferred: russh::Preferred {
                 kex: &[russh::kex::NONE],
                 key: &[russh_keys::key::NONE],
                 cipher: &[russh::cipher::NONE],
                 mac: russh::Preferred::DEFAULT.mac,
-                compression: russh::Preferred::COMPRESSED.compression,
+                compression: &["none"],
             },
             ..Default::default()
         };
@@ -595,6 +610,7 @@ impl Server {
 
         let config = self.config.clone();
         tokio::spawn(async move {
+            log::debug!("starting to serve host relay client session");
             let session = match russh::server::run_stream(config, rw, server_session).await {
                 Ok(s) => s,
                 Err(e) => {
@@ -608,6 +624,7 @@ impl Server {
                 return Ok(()); // session closed
             }
 
+            log::debug!("host relay client session successfully authed");
             let mut known_ports: PortMap = HashMap::new();
             tokio::pin!(session);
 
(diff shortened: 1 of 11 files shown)


connor4312:
russh is published, so this should be ready to go. Not sure who has the expertise to give this a good review, maybe @brettcannon but he's quite busy.

connor4312:
russh is published, so this should be ready to go

rs/src/connections/host_relay.rs
```diff
@@ -0,0 +1,1052 @@
...
+                log::info!("Error connecting forwarding to {}, {}", addr, e);
+                conn.close().await;
+                continue;
+            }
+        };
+
+        log::debug!("Forwarded port to {}", addr);
+
+        tokio::spawn(async move {
+            let mut read_buf = [0u8; 1024 * 64];
```
wesleywiser, line 685: I'm a bit concerned about allocating 64k of stack. (There's been some discussion of how we want to tune the unstable `large_assignment` lint [here](https://github.com/rust-lang/rust/issues/83518#issuecomment-1072779300) and 64k is called out as "huge")

I would suggest using `vec![0u8; 1024 * 64].into_boxed_slice()` to get a `Box<[u8]>` or something like [this](https://stackoverflow.com/a/72504064/124538) if you want a `Box<[u8; 1024 * 64]>`.

rs/src/connections/io.rs
```diff
@@ -0,0 +1,40 @@
...
+    /// Removes any data stored in the read buffer
+    pub fn take_data(&mut self) -> Option<Vec<u8>> {
+        self.0.take()
+    }
+
+    /// Writes as many bytes as possible to the readbuf, stashing any extra.
+    pub fn put_data(
+        &mut self,
+        target: &mut tokio::io::ReadBuf<'_>,
+        bytes: &[u8],
```
wesleywiser, line 21: It might make sense to take a `Message` by value here and move the `Message` from the caller. Doing that, you'd then put `Message` in `ReadBuffer` instead of the `Vec<u8>` and hat way, you don't have to copy the part of the message that didn't send on line 34. If the message is very large and the `ReadBuf` is very small, you could end up copying the same tail of the message over and over in the `AsyncRWWebSocket::poll_read` call site. 

This suggestion is really just about the perf aspect; I think the design here is fine so please ignore until you're working on perf.

rs/src/connections/ws.rs
```diff
@@ -0,0 +1,330 @@
...
+        .expect("expected to make req");
+
+        tokio::spawn(async move {
+            let (cnx, _) = echo_server.accept().await.expect("expect client");
+            accept_echo_server_connection(cnx).await;
+        });
+
+        let input_len = 1024 * 1024;
+        let mut input = Vec::new();
+        input.reserve(input_len);
```
wesleywiser, line 280: Nit: `Vec::with_capacity(input_len)`

wesleywiser (dismissed):
lgtm 😄

rs/src/connections/host_relay.rs
```diff
@@ -0,0 +1,1085 @@
...
+type PortMap = HashMap<u32, mpsc::UnboundedSender<ForwardedPortConnection>>;
+
+/// The HostRelay can host connections via the tunneling service. After
+/// creating it, you will generally want to run `connect()` to create a new
+/// a new connection.
+///
+/// Note that, while ports can be added and remove dynamically from running
+/// tunnels via the appropriate methods on the HostRelay, no ports will be
+/// hosted until those methods are called.
+pub struct HostRelay {
```
jasongin, line 48: I'm not sure about this name. In the tunneling terminology we use, the "relay" is the cloud service that relays data between host and client.

In other SDKs we call this thing `TunnelRelayTunnelHost`, where "relay" is describing the specific type of tunnel host. (Just `RelayTunnelHost` might be simpler, but there used to be another host type that used Live Share relays; that's obsolete now.) In the future there will be other host classes, like a `LocalNetworkTunnelHost` that accepts direct connections on a local network without going through a relay.

rs/src/connections/host_relay.rs
```diff
@@ -0,0 +1,1085 @@
...
+    ) -> Result<
+        (
+            russh::client::Handle<Client>,
+            mpsc::UnboundedReceiver<ChannelOp>,
+        ),
+        russh::Error,
+    > {
+        let config = russh::client::Config {
+            anonymous: true,
+            window_size: 1024 * 1024 * 64,
```
jasongin, line 356: The window size for the outer SSH session (the SSH client session between host and relay) should be larger than the window size for the inner SSH sessions (the SSH server sessions between clients and hosts), to prevent a single client from filliing the host's window and thereby blocking other clients. In https://github.com/microsoft/dev-tunnels/pull/135, I used 5 MB for the outer session and 1 MB (default) for the inner sessions; the relay services is also updated to use 5 MB for the host session.

Eventually this protocol will change to avoid layering SSH, and then the window size will be less important.

jasongin:
It would be nice to also have the client side of relay connections implemented. I know it's not currently needed but it would make end-to-end testing easier, and it's much simpler than the host side anyway.

This doesn't need to block the current PR.

rs/src/connections/host_relay.rs
```diff
@@ -0,0 +1,1085 @@
...
+                reason: "failed to remove port from tunnel",
+            })?;
+
+        self.ports_tx.send_modify(|v| {
+            v.remove(&(port_number as u32));
+        });
+
+        Ok(())
+    }
+
```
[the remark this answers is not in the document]
connor4312: > that causes the host to retrieve the updated port list from the service and add/remove forwarded ports accordingly

I don't think I would want this to be one-step like that in the Rust SDK. In the SDK, we give clients the option of handling port traffic directly rather than forwarding everything over TCP. We take advantage of this in the VS Code CLI. As a results, ports always get started or ended by the consumer telling the SDK how to handle the port. I think this isn't something that other SDKs support, which may impact how we think about the API.

From my understanding of your comment, in pseudo code adding a port in Rust would be something like:

```ts
mgmt.add_port(port)
relay.refresh_ports()
relay.tcp_forward_port(port) // or something like relay.get_port_stream
```

It seems like an `add_port`-esque API is more ergonomic, turning those three calls into one. The relay could also refresh automatically if the host asks for a port that it doesn't know about--though it's still another failure case in the SDK's API which isn't present with the current `add_port` method.

Regardless  of the above, if we support client requests for a port, we'd probably need some event/handler to be given to the RelayTunnelHost to figure out how to handle it.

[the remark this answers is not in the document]
jasongin: Other SDKs give _tunnel client_ applications the option to handle port connection streams directly, but currently assume _tunnel hosts_ are always forwarding TCP sockets. I agree that it is good to give hosting applications that flexibility also, as long as the more common case of forwarding TCP sockets is still simple enough.

> if we support client requests for a port, we'd probably need some event/handler to be given to the RelayTunnelHost to figure out how to handle it

Yes, we have scenarios in which something other than the host application is in control of which ports get forwarded. That was the reason for adding the refresh-ports API: to enable any authorized client to use the management API to add/remove ports. (Eventually I would like to eliminate the need for the explicit call to refresh-ports and replace it with some kind of message from the relay service that gets sent to the host automatically after any port changes via the management API.)

> It seems like an `add_port`-esque API is more ergonomic, turning those three calls into one.

My thinking was that it was better if the SDK did not offer multiple ways of doing the same thing, and easier for the host application developer to understand how it works if those calls are separate. But I'd be open to keeping `add_port`/`remove_port` as convenience methods if they document how they are equivalent of the other API calls.

> The relay could save a call by refreshing automatically if the host asks for a port that it doesn't know about

I assume you meant "if a _client_ asks for a port that that [the host] doesn't know about". I considered that as an alternative to the explicit refresh-ports message, but the problem is it doesn't work for _removed_ ports. If a client asks for a port that was removed, then the host won't know to refresh -- unless it refreshes before accepting every connection, but that would be slow. Also later there could be other properties on a port that affect how the host handles connections (for example protocol=udp).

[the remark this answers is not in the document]
jasongin: This issue doesn't need to block the PR. There will be more changes (across all the SDKs) anyway as we continue to refine and expand the APIs in this area. This is good for now.

jasongin (approved):

PR #131, opened 2022-09-12
Add default SSH command and default web uri

klvnraju:
Adding default ssh command and default web uri to tunnel endpoint for clients (for example CLI) to decide if we should show port number in tunnel web uri or ssh command


go/tunnels/tunnel_endpoint.go
```diff
@@ -1,76 +1,84 @@
-// Copyright (c) Microsoft Corporation.
-// Licensed under the MIT license.
-// Generated from ../../../cs/src/Contracts/TunnelEndpoint.cs
-
```
klvnraju, line 4: fix this

cs/src/Contracts/TunnelEndpoint.cs
```diff
@@ -61,6 +61,13 @@ public abstract class TunnelEndpoint
     [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
     public string? PortUriFormat { get; set; }
 
+    /// <summary>
+    /// Gets or sets the default URIs without the port information in the url
+    /// and the web forwarding forwards requests to this url to default port.
+    /// </summary>
+    [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
+    public string? DefaultWebUri { get; set; }
```
jasongin, line 69: I want to be careful about consistent naming here because this is a public API. This is a little inconsistent now because it has "web" while `PortUriFormat` does not. Also I'd prefer "tunnel" instead of "default" since I think it's unclear what "default" is referring to. So how about these 4 property names?

```
TunnelUri
PortUriFormat
TunnelSshCommand
PortSshCommandFormat
```

I also considered `TunnelWebUri` / `PortWebUriFormat`, but it's not worth a breaking change to `PortUriFormat`.

cs/src/Contracts/TunnelEndpoint.cs
```diff
@@ -61,6 +61,13 @@ public abstract class TunnelEndpoint
     [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
     public string? PortUriFormat { get; set; }
 
+    /// <summary>
+    /// Gets or sets the default URIs without the port information in the url
+    /// and the web forwarding forwards requests to this url to default port.
```
jasongin, line 66: The phrasing of this comment is  confusing. How about this?

`Gets or sets the URI where a web client can connect to the default port of the tunnel.`

cs/src/Contracts/TunnelEndpoint.cs
```diff
@@ -69,6 +75,13 @@ public abstract class TunnelEndpoint
     [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
     public string? PortSshCommandFormat { get; set; }
 
+    /// <summary>
+    /// Gets or sets the Ssh command where the Ssh Gateway can connect to the default ssh port
```
jasongin, line 79: Should this reference "ssh client" rather than "ssh gateway"?

jasongin (dismissed):

cs/src/Contracts/TunnelEndpoint.cs
klvnraju: yes, that is correct. Fixing it.

jasongin (approved):

PR #134, opened 2022-09-15
Add get ssh keys api in management client

klvnraju:
Add management client to `tunnelid/sshkey` api endpoint. This returns the ssh public key that has to be added to the `authorized_keys` file for the Ssh Gateway scenario. This is the key used by tunnels service to connect to the shared ssh port.


cs/src/Contracts/TunnelSshKeyResponse.cs
```diff
@@ -0,0 +1,19 @@
...
+/// <summary>
+/// Response for SshKey endpoint.
+/// </summary>
+public class TunnelSshKeyResponse
+{
+    /// <summary>
+    /// Gets or sets the ssh key for a tunnel.
+    /// </summary>
+    [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
+    public string? SshKey { get; set; }
```
jasongin, line 18: What if we ever want to rotate the key, should there be a primary and secondary (or array) here?

cs/src/Management/ITunnelManagementClient.cs
```diff
@@ -345,5 +345,18 @@ public interface ITunnelManagementClient : IDisposable
         /// <param name="cancellation">Cancellation token.</param>
         /// <returns>Array of <see cref="ClusterDetails"/></returns>
         Task<ClusterDetails[]> ListClustersAsync(CancellationToken cancellation = default);
+
+        /// <summary>
+        /// Gets the ssh public key for an ssh client to connect to a shared ssh server.
```
jasongin, line 350: Please explain better what scenario the key is for and what one would do with the key (add to `authorized_keys`).

cs/src/Contracts/TunnelSshKeyResponse.cs
jasongin, line 18: How about naming it `SshGatewayPublicKey`?

cs/src/Contracts/TunnelSshKeyResponse.cs
```diff
@@ -0,0 +1,19 @@
...
+// Licensed under the MIT license.
+// </copyright>
+using System.Text.Json.Serialization;
+
+namespace Microsoft.VsSaaS.TunnelService.Contracts;
+
+/// <summary>
+/// Response for SshKey endpoint.
+/// </summary>
+public class TunnelSshKeyResponse
```
jasongin, line 12: Could the SSH gateway public key be a property on `TunnelRelayTunnelEndpoint`, instead of a separate API + contract? It is related to the relay endpoint in that the relay is acting as the SSH gateway. And there are other SSH properties on the relay endpoint object already.

jasongin: ... but only if it does not add extra time/cost (such as a keyvault query) to operations that retrieve the endpoint.

PR #137, opened 2022-09-21
Add Java refresh ports tests

jessetrinity:
* Respond to `tcpip-forward` request with success if the requested port is already in the forwarded ports collection. 
* Add test for adding and removing ports.

### DIFF
--- a/java/src/main/java/com/microsoft/tunnels/connections/TcpipForwardRequestHandler.java
+++ b/java/src/main/java/com/microsoft/tunnels/connections/TcpipForwardRequestHandler.java
@@ -55,14 +55,28 @@ class TcpipForwardRequestHandler extends AbstractConnectionServiceRequestHandler
         connectionService.getForwarder(), "No TCP/IP forwarder");
 
     String address = buffer.getString();
-    int requested = buffer.getInt();
-    int port = requested;
+    int requestedPort = buffer.getInt();
+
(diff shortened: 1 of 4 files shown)


jfullerton44 (approved):

java/src/main/java/com/microsoft/tunnels/connections/TcpipForwardRequestHandler.java
```diff
@@ -55,14 +55,28 @@ public Result process(
         connectionService.getForwarder(), "No TCP/IP forwarder");
 
     String address = buffer.getString();
-    int requested = buffer.getInt();
-    int port = requested;
+    int requestedPort = buffer.getInt();
+
+    // Ignore the request if the port is already in the collection.
```
jasongin, line 60: I thought this shouldn't ever happen if the host is operating correctly. I'd like to understand why it is happening, before ignoring it.

jasongin: (As I recall, the client was _not_ making current requests, so I don't think that is the explanation.)

jessetrinity: I've updated the PR to only add refresh ports tests.

jasongin (approved):

PR #138, opened 2022-09-22
Include ssh gateway public key in UpdateTunnelEndpoint

klvnraju:
Send additional parameter to include ssh gateway public key when calling UpdateTunnelEndpoint api if there is atleast one ssh port


cs/src/Management/TunnelManagementClient.cs
```diff
@@ -921,13 +921,19 @@ public void Dispose()
             Requires.NotNull(endpoint, nameof(endpoint));
             Requires.NotNullOrEmpty(endpoint.HostId!, nameof(TunnelEndpoint.HostId));
 
+            string? query = null;
+            if (tunnel.Ports != null && tunnel.Ports.Where((p) => p.Protocol == TunnelProtocol.Ssh).Any())
+            {
+                query = "includeSshGatewayPublicKey=true";
```
IlyaBiryukov, line 927: Can `includeSshGatewayPublicKey` be a string constant in a contract? Then tunnel service may also use that constant.

IlyaBiryukov (dismissed):

cs/src/Management/TunnelManagementClient.cs
```diff
@@ -921,13 +921,19 @@ public void Dispose()
             Requires.NotNull(endpoint, nameof(endpoint));
             Requires.NotNullOrEmpty(endpoint.HostId!, nameof(TunnelEndpoint.HostId));
 
+            string? query = null;
+            if (tunnel.Ports != null && tunnel.Ports.Where((p) => p.Protocol == TunnelProtocol.Ssh).Any())
```
jasongin, line 925: nit: `.Any(lambda)` instead of `.Where(lambda).Any()`

jasongin:
An alternative approach could be to add a `TunnelRequestOptions.AdditionalQueryParameters`, similar to `AdditionalHeaders`.

jasongin:
Why is the option only on update operation and not get operation? Can you explain the scenario where this is used?

klvnraju:
There is no get for endpoints api.
The tunnel CLI uses this update api while hosting a tunnel and uses the output to show relay url/ssh command.

cs/src/Management/TunnelManagementClient.cs
```diff
@@ -921,13 +921,19 @@ public void Dispose()
...
+            {
+                query = "includeSshGatewayPublicKey=true";
+            }
```
klvnraju, line 928: Do this in typescript SDK as well

jasongin:
> There is no get for endpoints api.

The get-tunnel response includes endpoints.

klvnraju:
The get-tunnel implementation has port information to determine if there is an ssh port and hence does not require this query parameter. (Only exception is if we call get-tunnel with a non-ssh port number and there are other ports which are ssh.. but in that scenario, we can skip sending the key)

PR #141, opened 2022-09-27
Add AdditionalQueryParameters and include ssh gateway public key in UpdateTunnelEndpoint

klvnraju:
Add AdditionalQueryParameters for users of SDK to send additional query parameters.
Send additional parameter to include ssh gateway public key when calling UpdateTunnelEndpoint api if there is atleast one ssh port.

### DIFF
--- a/cs/src/Management/TunnelRequestOptions.cs
+++ b/cs/src/Management/TunnelRequestOptions.cs
@@ -43,7 +43,7 @@ namespace Microsoft.VsSaaS.TunnelService
         public IEnumerable<KeyValuePair<string, string>>? AdditionalHeaders { get; set; }
 
         /// <summary>
-        /// Gets or sets additional headers to be included in the request.
+        /// Gets or sets additional query parameters to be included in the request.
         /// </summary>
         public IEnumerable<KeyValuePair<string, string>>? AdditionalQueryParameters { get; set; }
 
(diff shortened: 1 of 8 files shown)


cs/src/Management/TunnelRequestOptions.cs
```diff
@@ -42,6 +42,11 @@ public class TunnelRequestOptions
         /// </summary>
         public IEnumerable<KeyValuePair<string, string>>? AdditionalHeaders { get; set; }
 
+        /// <summary>
+        /// Gets or sets additional headers to be included in the request.
```
jasongin, line 46: Update the comment.

ts/src/management/tunnelManagementHttpClient.ts
```diff
@@ -606,6 +606,22 @@ export class TunnelManagementHttpClient implements TunnelManagementClient {
             tunnelPath = `${tunnelsApiPath}/${tunnel.name}`;
         }
 
+        var additionalQuery = '';
+        if (options?.additionalQueryParameters) {
+            for (let [paramName, paramValue] of Object.entries(options.additionalQueryParameters)) {
+                additionalQuery = `${paramName}=${paramValue}`;
```
jasongin, line 612: I think this was meant to be something like:
```TypeScript
addtionalQuery += `&${paramName}=${paramValue}`;
```

jasongin (approved):

jasongin (approved):

PR #142, opened 2022-09-29
Use open source ssh package and publish to external feed

jessetrinity:
* Change `@vs/vs-ssh` dependency to `@microsoft/dev-tunnels-ssh` package.
* Rename `@vs/tunnels-*` packages to `@microsoft/tunnels-*` and publish to npm.
* Rename `Microsoft.VsSaas.TunnelService` to `Microsoft.DevTunnel` and publish to NuGet.

We should probably rename the `@microsoft/tunnels-*` packages to `@microsoft/dev-tunnels-*` to match the others.


cs/src/Contracts/TunnelService.Contracts.csproj
```diff
@@ -1,8 +1,8 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <RootNamespace>Microsoft.VsSaaS.TunnelService.Contracts</RootNamespace>
-    <AssemblyName>Microsoft.VsSaaS.TunnelService.Contracts</AssemblyName>
+    <RootNamespace>Microsoft.DevTunnels.TunnelService.Contracts</RootNamespace>
```
jasongin, line 4: Should the namespace be just `Microsoft.DevTunnels.Contracts`? The `TunnelService` part seems redundant, and maybe a little confusing since this is not service code.

Same for the Management and Connections project.

jessetrinity: Updated.

cs/src/Connections/TunnelService.Connections.csproj
```diff
@@ -1,8 +1,8 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <AssemblyName>Microsoft.VsSaaS.TunnelService.Connections</AssemblyName>
-    <RootNamespace>Microsoft.VsSaaS.TunnelService</RootNamespace>
+    <AssemblyName>Microsoft.DevTunnels.Connections</AssemblyName>
+    <RootNamespace>Microsoft.DevTunnels</RootNamespace>
```
jasongin, line 5: The namespace here, and for all the classes in this project, should probably include the `.Connections` suffix.

When I suggested the namespace change I was thinking it was already like that, but I see from the diff it was not.

jasongin: And same for `Management`. Oddly the `Contracts` project already has the suffix.

Probably I made this mistake when originally splitting into the 3 assemblies, and never thought about it afterward.

cs/src/Connections/TunnelService.Connections.csproj
```diff
@@ -1,8 +1,8 @@
 <Project Sdk="Microsoft.NET.Sdk">
```
jasongin, line 1: I'd suggest naming the `.csproj` files also: `TunnelService.Connections.csproj` -> `DevTunnels.Connections.csproj`, and similar.

Or we can wait and do further renaming in a separate change.

jasongin (approved):

cs/src/Connections/TunnelService.Connections.csproj
jessetrinity: Updated.

cs/src/Connections/TunnelService.Connections.csproj
jessetrinity: Updated.

cs/src/Connections/TunnelService.Connections.csproj
jasongin: I was suggesting to update the namespace in all the code files also. Or we can do that separately.

jessetrinity: Right🤦

I think that should go with this change to avoid having to update the imports twice wherever this is consumed.

jasongin (approved):

PR #143, opened 2022-10-03
Add http request rate to tunnel status

jfullerton44:
This adds the http request rate to the tunnel status. This will be used to http request rate limiting in the service

### DIFF
--- a/cs/src/Contracts/TunnelStatus.cs
+++ b/cs/src/Contracts/TunnelStatus.cs
@@ -96,7 +96,8 @@ public class TunnelStatus
     public RateStatus? ApiUpdateRate { get; set; }
 
     /// <summary>
-    /// Gets or sets the current value and limit for the rate of http requests to the tunnel.
+    /// Gets or sets the current value and limit for the rate of http requests to the tunnel
+    /// when web forwarding.
     /// </summary>
     [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
     public RateStatus? HttpRequestRate { get; set; }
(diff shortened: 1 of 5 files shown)


IlyaBiryukov (approved):

cs/src/Contracts/TunnelStatus.cs
```diff
@@ -94,4 +94,10 @@ public class TunnelStatus
     /// </summary>
     [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
     public RateStatus? ApiUpdateRate { get; set; }
+
+    /// <summary>
+    /// Gets or sets the current value and limit for the rate of http requests to the tunnel.
```
derekbekoe, line 99: Should we mention "web forwarding" so it's clear this isn't management plane operations on the tunnel? Same for the other language comments.

jfullerton44: good point, added

jasongin: Eventually I'd like this to include HTTP requests through desktop / SDK client forwarding also. But that will require  changing the protocol, as I've mentioned before.

jasongin (approved):

jfullerton44:
This actually might not be needed since the tunnel port status has this filed

jasongin:
> This actually might not be needed since the tunnel port status has this filed

Good point. Yes, my original intention had been to limit the HTTP request rate per-port.

IlyaBiryukov (approved):

PR #149, opened 2022-10-10
On 429 when connecting, limit retries and space out retries more

jfullerton44:
This changes the behavior for client connection so that if a 429 is returned it will only retry 4 times, increasing the delay 4x between each retry


cs/src/Connections/RelayTunnelConnector.cs
```diff
@@ -121,7 +121,15 @@ public RelayTunnelConnector(IRelayClient relayClient)
                         case HttpStatusCode.TooManyRequests:
                             errorDescription = "Rate limit exceeded (429). Too many requests in a given amount of time.";
                             exception = new TunnelConnectionException(errorDescription, wse);
+                            if (attempt > 4)
+                            {
+                                throw exception;
+                            }
+
+                            attemptDelayMs <<= 1;
```
IlyaBiryukov, line 129: Should we check `if (attemptDelayMs < RetryMaxDelayMs)` ?

cs/src/Connections/RelayTunnelConnector.cs
```diff
@@ -121,7 +121,15 @@ public RelayTunnelConnector(IRelayClient relayClient)
                         case HttpStatusCode.TooManyRequests:
                             errorDescription = "Rate limit exceeded (429). Too many requests in a given amount of time.";
                             exception = new TunnelConnectionException(errorDescription, wse);
```
IlyaBiryukov, line 123: It would be great to have its own exception type for the rate limit exception.
Ideally, with additional rate limit properties from the response headers. Unfortunately, there is no way to get response headers from web socket error in 6.0 (there is a way in 7.0 though)

ts/src/connections/relayTunnelConnector.ts
```diff
@@ -218,7 +218,12 @@ export class RelayTunnelConnector implements TunnelConnector {
 
                         case 'error.tooManyRequests':
                             errorDescription = `Rate limit exceeded${statusCodeText}. Too many requests in a given amount of time.`;
-                            continue;
+                            if (attempt > 4) {
+                                throwError(errorDescription);
+                            }
+
+                            attemptDelayMs = attemptDelayMs << 1;
```
IlyaBiryukov, line 225: Same here.

IlyaBiryukov (approved):

ts/src/connections/relayTunnelConnector.ts
jfullerton44: updated

cs/src/Connections/RelayTunnelConnector.cs
jfullerton44: Created this issue to tracking doing that in the future https://github.com/microsoft/basis-planning/issues/541

IlyaBiryukov (approved):

PR #151, opened 2022-10-18
Add header to skip antiphishing page 

jfullerton44:
This adds a header that can be sent with requests that will cause the service to return the skipPhishing cookie and skip the antiphishing page during webforwarding

### DIFF
--- a/cs/src/Contracts/TunnelHeaderNames.cs
+++ b/cs/src/Contracts/TunnelHeaderNames.cs
@@ -27,4 +27,9 @@ public static class TunnelHeaderNames
     /// Github Ssh public key which can be used to validate if it belongs to tunnel's owner.
     /// </summary>
     public const string XGithubSshKey = "X-Github-Ssh-Key";
+
+    /// <summary>
+    /// Header that will skip the antiphishing page when connection to a tunnel through web forwarding.
+    /// </summary>
+    public const string XTunnelSkipPhishing = "X-Tunnel-Skip-Phishing-Page";
 }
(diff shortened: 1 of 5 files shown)


cs/src/Contracts/TunnelHeaderNames.cs
```diff
@@ -27,4 +27,9 @@ public static class TunnelHeaderNames
     /// Github Ssh public key which can be used to validate if it belongs to tunnel's owner.
     /// </summary>
     public const string XGithubSshKey = "X-Github-Ssh-Key";
+
+    /// <summary>
+    /// Header that will skip the antiphishing page when connection to a tunnel through web forwarding.
+    /// </summary>
+    public const string XTunnelSkipPhishing = "X-Tunnel-Skip-Phishing-Page";
```
IlyaBiryukov, line 34: `XTunnelSkipPhishingPage`

IlyaBiryukov (approved):

cs/src/Contracts/TunnelHeaderNames.cs
jasongin, line 34: The name sounds a little like it's skipping a malicious page. Maybe it should be `X-Tunnel-Skip-AntiPhishing-Page` or `X-Tunnel-Skip-Interstitial -Page`?

jfullerton44: updated

jfullerton44: updated

jasongin (approved):

PR #152, opened 2022-10-19
Add HostHeader tunnel option

IlyaBiryukov:
To let tunnels change the `Host` request header in web-forwarding, we need to pass this header to the service.


jfullerton44 (dismissed):

IlyaBiryukov:
More changes are coming. We'll need an option to keep the host header "as is", and, maybe similar options for other request headers, like `Origin`.

jfullerton44 (approved):

PR #154, opened 2022-10-28
Report more bandwidth details in tunnel status

jasongin:
- Extend `TunnelStatus` properties to track `UploadRate`, `UploadTotal`, `DownloadRate`, `DownloadTotal`.
 - Add `ToString()` methods in C# `ResourceStatus` and `RateStatus` classes.

### DIFF
--- a/cs/src/Contracts/RateStatus.cs
+++ b/cs/src/Contracts/RateStatus.cs
@@ -27,4 +27,47 @@ public class RateStatus : ResourceStatus
     /// </summary>
     [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
     public long? ResetTime { get; set; }
+
+    /// <inheritdoc/>
+    public override string ToString()
+    {
+        var count = base.ToString();
+        if (PeriodSeconds == null)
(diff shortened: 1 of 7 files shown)


cs/src/Contracts/RateStatus.cs
```diff
@@ -27,4 +27,47 @@ public class RateStatus : ResourceStatus
...
+        if (PeriodSeconds == null)
+        {
+            return count;
+        }
+
+        if (PeriodSeconds.Value == 1)
+        {
+            return count + "/s";
+        }
+        else if (PeriodSeconds.Value == 60)
```
IlyaBiryukov, line 44: nit: all these `else` are not necessary because there is `return` inside `if` branches.

IlyaBiryukov (approved):

cs/src/Contracts/RateStatus.cs
jasongin: Sure, but I like the `else` keywords anyway because the clauses are all related. And without `else` the style rules would force an extra blank line after each `}`.

PR #155, opened 2022-10-28
Add Limit Param to List Operation

jfullerton44:
This adds a param to the tunnel list command to limit the number of tunnels returned. This can speed up queries to control plane when searching for single tunnels by tags

### DIFF
--- a/cs/src/Management/TunnelRequestOptions.cs
+++ b/cs/src/Management/TunnelRequestOptions.cs
@@ -116,7 +116,7 @@ namespace Microsoft.DevTunnels.Management
         /// <summary>
         /// Limits the number of tunnels returned when searching or listing tunnels.
         /// </summary>
-        public uint Limit { get; set; }
+        public uint? Limit { get; set; }
 
         /// <summary>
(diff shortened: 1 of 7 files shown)


cs/src/Management/TunnelRequestOptions.cs
```diff
@@ -113,6 +113,11 @@ public class TunnelRequestOptions
         /// </summary>
         public bool ForceRename { get; set; }
 
+        /// <summary>
+        /// Limits the number of tunnels returned when searching or listing tunnels.
```
IlyaBiryukov, line 117: Are we going to have a default limit if the request option is not provided?
Also, are we going to limit the `Limit` itself?

IlyaBiryukov (approved):

cs/src/Management/TunnelRequestOptions.cs
jfullerton44: If limit is not provided then all results will be returned, if the limit is larger than the number of results then it will just return all results.

PR #156, opened 2022-10-31
Add attribute based validation to contracts

jasongin:


cs/src/Contracts/TunnelAccessControlEntry.cs
```diff
@@ -79,6 +87,7 @@ public TunnelAccessControlEntry()
         /// For anonymous ACEs, this value is null.
         /// </remarks>
         [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
+        [StringLength(Providers.MaxLength)]
```
IlyaBiryukov, line 90: Should we limit to known providers with a regular expression?

cs/src/Contracts/TunnelAccessControlEntry.cs
```diff
@@ -127,12 +136,15 @@ public TunnelAccessControlEntry()
         /// with any other types of ACEs.
         /// </remarks>
         [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
+        [StringLength(OrganizationMaxLength)]
```
IlyaBiryukov, line 139: Are there valid or invalid chars defined for this property,? E.g. `<,` `&`, `/`, `\`, `"`, `>` are most likely invalid. 
Limiting these string contents and prohibiting invalid input will help mitigate stored or reflected XSS ([Types of XSS](https://owasp.org/www-community/Types_of_Cross-Site_Scripting)]

Same for other properties that don't have a regular expression validation.
Not blocking for this PR, but still a good thing to have, defense in depth.

cs/src/Contracts/TunnelAccessControlEntry.cs
```diff
@@ -141,6 +153,8 @@ public TunnelAccessControlEntry()
         /// <remarks>
         /// These must be one or more values from <see cref="TunnelAccessScopes" />.
         /// </remarks>
+        [MaxLength(MaxScopes)]
+        [ArrayStringLength(TunnelAccessScopes.MaxLength)]
```
IlyaBiryukov, line 157: It would be great to have a regex validating known scopes and their derivatives (i.e. multiple scopes delimited by space)

cs/src/Contracts/TunnelOptions.cs
```diff
@@ -12,6 +13,8 @@ namespace Microsoft.DevTunnels.Contracts
     /// </summary>
     public class TunnelOptions
     {
+        private const int HeaderMaxLength = 500;
```
IlyaBiryukov, line 16: This is 255 + 1 + 5, max DNS name length (255) + 1 for ':' + 5 for '65535', max port length.

cs/src/Contracts/TunnelOptions.cs
```diff
@@ -52,6 +56,7 @@ public class TunnelOptions
         /// The option is ignored if IsOriginHeaderUnchanged is true.
         /// </summary>
         [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingDefault)]
+        [StringLength(HeaderMaxLength)]
```
IlyaBiryukov, line 59: The origin is {scheme:}//{host}:{port}
This also includes a scheme, so the max length is 255 + 1 + 5 + 8 ,
where max DNS name length is 255, + 1 for ':', + 5 for '65535', max port length, and +8 for 'https://', and  +1 for trailing '/'

cs/src/Contracts/TunnelPort.cs
```diff
@@ -4,109 +4,124 @@
...
+    [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
+    [RegularExpression(TunnelNamePattern)]
+    [StringLength(TunnelNameMaxLength)]
+    public string? Name { get; set; }
+
+    /// <summary>
+    /// Gets or sets the optional description of the port.
+    /// </summary>
+    [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
+    [StringLength(DescriptionMaxLength)]
```
IlyaBiryukov, line 63: Do we have disallowed characters for tunnel descriptions? Perhaps we shouldn't allow html-control characters <, &, >, and maybe also ", ', /, and \\

cs/src/Contracts/TunnelPort.cs
```diff
@@ -4,109 +4,124 @@
...
+
+    /// <summary>
+    /// Gets or sets the protocol of the tunnel port.
+    /// </summary>
+    /// <remarks>
+    /// Should be one of the string constants from <see cref="TunnelProtocol"/>.
+    /// </remarks>
+    [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
+    [StringLength(TunnelProtocol.MaxLength)]
+    public string? Protocol { get; set; }
```
IlyaBiryukov, line 83: Add a regex validation with all know valid protocols

IlyaBiryukov (approved):

cs/src/Contracts/TunnelAccessControlEntry.cs
jasongin: Yes, `Organization` and `Subject` here should use the same validation regex as in `TunnelAccessSubject`. I forgot to apply it here.

cs/src/Contracts/TunnelAccessControlEntry.cs
jasongin: For that to be maintainable, the pattern needs to be generated from the list of all scopes. But `ValidateRegularExpressionAttribute` requires a compile-time string constant... unless I extend the attribute to take a method name or something and then use reflection, or use a source-generator, but either of those approaches would be messy. I think it's good enough to just do length validation here, and leave the rest of the validation to the code in the service.

cs/src/Contracts/TunnelAccessControlEntry.cs
jasongin: Not worth it in my opinion - see my comment about a scopes regex.

cs/src/Contracts/TunnelPort.cs
jasongin: How much value would there be in that, really? Can a fuzzer even generate values that match a regex?

It could be cleaner to write a custom validation attribute that checks if the value matches a list of strings. But the service already effectively does that anyway, so I think that would not be very useful either.

cs/src/Contracts/TunnelOptions.cs
jasongin: I think there's no need to be so exact. My intention for this validation is to prevent unreasonably large inputs, so I'm just going to use 300 for both headers for now. You can change that or add regex validation for these later.

cs/src/Contracts/TunnelPort.cs
jasongin: Depending on the context there are many characters that can be dangerous, and I think it's unreasonable to block or strip all of them in a description. Since we cannot strictly validate this property, the proper mitigation is to ensure that anywhere the description is rendered that it is properly encoded/escaped at that point.

cs/src/Contracts/TunnelAccessControlEntry.cs
jasongin: Fixed.

IlyaBiryukov (approved):

jfullerton44 (approved):

PR #160, opened 2022-11-03
Move cs and ts PR build to GH actions

jessetrinity:


.github/workflows/typescript-pr-build.yml
```diff
@@ -0,0 +1,36 @@
+name: TypeScript PR Build
+
+on:
+  pull_request:
```
OzVazquez, line 4: Yy

.github/workflows/typescript-pr-build.yml
```diff
@@ -0,0 +1,36 @@
...
+    branches: [ main ]
```
OzVazquez, line 5: ViVisual

OzVazquez: Yh

PR #163, opened 2022-11-10
Add ssh gateway rate limit per ip address to tunnel status

klvnraju:
Required for https://devdiv.visualstudio.com/OnlineServices/_git/tunnels/pullrequest/434563

### DIFF
--- a/cs/src/Contracts/TunnelStatus.cs
+++ b/cs/src/Contracts/TunnelStatus.cs
@@ -135,4 +135,11 @@ public class TunnelStatus
     /// </summary>
     [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
     public RateStatus? ApiUpdateRate { get; set; }
+
+    /// <summary>
+    /// Gets or sets the current value and limit for the rate of Ssh Gateway connections
+    /// per IP Address.
+    /// </summary>
+    [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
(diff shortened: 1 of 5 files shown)


cs/src/Contracts/TunnelStatus.cs
```diff
@@ -135,4 +135,11 @@ public class TunnelStatus
     /// </summary>
     [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
     public RateStatus? ApiUpdateRate { get; set; }
+
+    /// <summary>
+    /// Gets or sets the current value and limit for the rate of Ssh Gateway connections
+    /// per IP Address.
+    /// </summary>
+    [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
+    public ulong? SshGatewayRateLimitPerIPAddress { get; set; }
```
jasongin, line 144: Is this limit also per-tunnel? Or per-IP address for all tunnels?

If it is per-tunnel, I'm not convinced that IP address rate-limiting is necessary: it would seem sufficient to have a rate limit on SSH connections for the tunnel.

If the limit is not per-tunnel, then the rate limit can be too easily reached when there are many users behind the same proxy IP address.

klvnraju: It is rate limit per IP address and not per tunnel. And your point is valid for users behind same firewall.

klvnraju:
Abandoning this PR as we are not using this as tunnel status as this is not tunnel specific

PR #173, opened 2022-12-08
Add inspectionEnabled to tunnel contracts

jfullerton44:
### Changes proposed: 
- Adds inspectionEnabled field to all tunnel contracts

### Other Tasks:

### DIFF
--- a/cs/src/Contracts/Tunnel.cs
+++ b/cs/src/Contracts/Tunnel.cs
@@ -128,6 +128,12 @@ public class Tunnel
     public TunnelPort[]? Ports { get; set; }
 
     /// <summary>
+    /// Gets or sets if inspection is enabled for the tunnel.
+    /// </summary>
+    [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
+    public bool? InspectionEnabled { get; set; }
+
+    /// <summary>
     /// Gets or sets the time in UTC of tunnel creation.
     /// </summary>
     [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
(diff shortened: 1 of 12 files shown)


cs/src/Contracts/Tunnel.cs
```diff
@@ -127,6 +127,12 @@ public Tunnel()
     [MaxLength(TunnelMaxPorts)]
     public TunnelPort[]? Ports { get; set; }
 
+    /// <summary>
+    /// Gets or sets if inspection is enabled for the tunnel.
+    /// </summary>
+    [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
+    public bool? InspectionEnabled { get; set; }
```
IlyaBiryukov, line 134: nit: `IsInspectionEnabled`

jfullerton44: updated

cs/src/Contracts/Tunnel.cs
```diff
@@ -127,6 +127,12 @@ public Tunnel()
     [MaxLength(TunnelMaxPorts)]
     public TunnelPort[]? Ports { get; set; }
 
+    /// <summary>
```
IlyaBiryukov, line 130: Why set it on the tunnel object and not on `TunnelOptions`? Tunnel options also apply to tunnel ports, and so can be controlled with more granularity.

IlyaBiryukov (approved):

cs/src/Contracts/Tunnel.cs
derekbekoe: Should this be inside of `TunnelPort`?
Or you're thinking that inspection should be enabled on a Tunnel and they all Ports on that tunnel get auto-inspected?

IlyaBiryukov: The tunnel options are stored as part of the tunnel and tunnel port objects. They're not a separate object.

IlyaBiryukov (approved):

PR #177, opened 2022-12-13
Implement suspendRead and resumeRead in Java

jfullerton44:
### Changes proposed: 
- Implements the suspendRead and ResumeRead functions in WebSocketSession using a semaphore

### DIFF
--- a/java/src/main/java/com/microsoft/tunnels/websocket/WebSocketSession.java
+++ b/java/src/main/java/com/microsoft/tunnels/websocket/WebSocketSession.java
@@ -18,7 +18,6 @@ import org.apache.sshd.common.io.IoWriteFuture;
 import org.apache.sshd.common.util.buffer.Buffer;
 import org.apache.sshd.netty.NettyIoService;
 import org.apache.sshd.netty.NettyIoSession;
-import org.apache.sshd.netty.NettySupport;
 
 public class WebSocketSession extends NettyIoSession {
   protected WebSocketConnectionHandler webSocketConnectionHandler;
--- a/java/src/main/java/com/microsoft/tunnels/websocket/WebSocketSession.java
+++ b/java/src/main/java/com/microsoft/tunnels/websocket/WebSocketSession.java
@@ -21,7 +21,7 @@ import org.apache.sshd.netty.NettyIoSession;
 
 public class WebSocketSession extends NettyIoSession {
   protected WebSocketConnectionHandler webSocketConnectionHandler;
-  protected Semaphore reading;
+  private Semaphore reading;
   /**
    * Creates a modified Netty session to handle websocket messages.
    */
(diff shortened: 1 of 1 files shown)


java/src/main/java/com/microsoft/tunnels/websocket/WebSocketSession.java
```diff
@@ -81,13 +85,13 @@ public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) throws E
...
-  public void suspendRead()  {
-    super.suspendRead();
+  public void suspendRead() {
+    this.reading.tryAcquire();
   }
 
   @Override
   public void resumeRead()  {
-    super.resumeRead();
+    this.reading.release();
```
IlyaBiryukov, line 94: What if `suspendRead` couldn't acquire the semaphore?

java/src/main/java/com/microsoft/tunnels/websocket/WebSocketSession.java
```diff
@@ -72,6 +74,8 @@ public void channelInactive(ChannelHandlerContext ctx) throws Exception {
 
   @Override
   public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception {
+    this.reading.acquire();
+    this.reading.release();
```
IlyaBiryukov, line 78: Should the release happen after `super.channelRead()`?

java/src/main/java/com/microsoft/tunnels/websocket/WebSocketSession.java
```diff
@@ -20,7 +21,7 @@
 
 public class WebSocketSession extends NettyIoSession {
   protected WebSocketConnectionHandler webSocketConnectionHandler;
-
+  protected Semaphore reading;
```
IlyaBiryukov, line 24: Can it be `private`?

java/src/main/java/com/microsoft/tunnels/websocket/WebSocketSession.java
jfullerton44: If it cant acquire the semaphore that means that suspendRead was already called. The tryAcquire will always run before Acquire

java/src/main/java/com/microsoft/tunnels/websocket/WebSocketSession.java
jfullerton44: Multiple reads are allowed at the same time, we just need to check if reads are blocked before starting

java/src/main/java/com/microsoft/tunnels/websocket/WebSocketSession.java
jfullerton44: updated

java/src/main/java/com/microsoft/tunnels/websocket/WebSocketSession.java
jfullerton44: if the semaphore is available when resumeRead is called it just returns immediately

IlyaBiryukov (approved):

PR #178, opened 2022-12-20
Restrict tunnel names ending in -inspect

jfullerton44:
Fixes issue where tunnels ending in -inspect can get confused with inspection service routing

### Changes proposed: 
- Adds restriction that tunnel names cannot end in -inspect as this will be used by the inspection urls
- If name is rejected, reason will be shown in CLI (separate PR)

### DIFF
--- a/cs/src/Contracts/TunnelConstraints.cs
+++ b/cs/src/Contracts/TunnelConstraints.cs
@@ -306,7 +306,7 @@ public static class TunnelConstraints
         {
             return false;
         }
-        if (tunnelName.EndsWith("-inspect"))
+        if (tunnelName.EndsWith("-inspect", StringComparison.OrdinalIgnoreCase))
         {
             return false;
         }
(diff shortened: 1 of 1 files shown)


cs/src/Contracts/TunnelConstraints.cs
```diff
@@ -306,6 +306,10 @@ public static bool IsValidTunnelName(string tunnelName)
         {
             return false;
         }
+        if (tunnelName.EndsWith("-inspect"))
```
IlyaBiryukov, line 309: Use overload with string comparison `OrdinalIgnoreCase`. The tunnel names are use in host name, and that is case-insensitive.

IlyaBiryukov (approved):

cs/src/Contracts/TunnelConstraints.cs
jfullerton44: added

IlyaBiryukov (approved):

PR #183, opened 2023-01-26
Dev/tasogawa/pdb

taiyosogawa:
This change separates our symbols from our published dlls (previously embedded) and adds pipeline steps to publish them to the appropriate locations. For reference, see this analogous PR to our SSH repo: https://github.com/microsoft/dev-tunnels-ssh/pull/28

### Changes proposed: 
- Removes the "embedded" debug type for the dll
- Adds a step to install the pdb2pdb converter on build machines, then convert portable pdbs to windows pdbs
- Uploads portable pdbs wrapped in a .snupkg to nuget.org and portable pdbs to the Microsoft symbol server
- Publishes the windows pdbs to the build pipline artifacts

I verified the steps by running the CI pipeline with the nuget.org step disabled, then queried the microsoft servers for the newly published symbols using symchk.

### Other Tasks:
- I updated our expiry tracker with an entry for the step that uploads symbols to the microsoft symbol server


cs/src/Contracts/DevTunnels.Contracts.csproj
```diff
@@ -9,11 +9,11 @@
...
     <UseNewtonsoftJson>false</UseNewtonsoftJson>
     <UseVisualStudioValidation>false</UseVisualStudioValidation>
-    <DebugType>embedded</DebugType>
     <GenerateDocumentationFile>True</GenerateDocumentationFile>
     <WarningsNotAsErrors>CS1591</WarningsNotAsErrors>
-    <!-- Include PDB files in NuGet package -->
-    <AllowedOutputExtensionsInPackageBuildOutputFolder>$(AllowedOutputExtensionsInPackageBuildOutputFolder);.pdb</AllowedOutputExtensionsInPackageBuildOutputFolder>
+    <!-- Create snupkg for debug symbols: https://learn.microsoft.com/en-us/nuget/create-packages/symbol-packages-snupkg#creating-a-symbol-package -->
+    <IncludeSymbols>true</IncludeSymbols>
+    <SymbolPackageFormat>snupkg</SymbolPackageFormat>
```
jasongin, line 16: Should these lines be moved to `cs/src/Directory.Build.props`?

jasongin (approved):

cs/src/Contracts/DevTunnels.Contracts.csproj
IlyaBiryukov: If we move these lines to `Directory.Build.props`, we should exclude the test projects from symbol publishing if we don't already do that.

IlyaBiryukov (approved):

cs/src/Contracts/DevTunnels.Contracts.csproj
jasongin: Test projects are under a different subdirectory (`cs/test`).

taiyosogawa: Ah, I like that idea - I've moved them to `Directory.Build.props` 😄

jasongin (approved):

PR #184, opened 2023-01-30
Specify nuget version 6.4.x in build pipeline

taiyosogawa:
Our build pipeline failed to publish the new .snupkg files to nuget.org because it was using an old version of the nuget CLI. This will install an appropriate version of nuget and prepend it to the build machine's PATH.

Here are some lines from our build pipeline output.

```
Detected NuGet version 4.1.0.2450 / 4.1.0
...

##[error]The nuget command failed with exit code(1) and error(System.AggregateException: One or more errors occurred. ---> System.Net.Http.HttpRequestException: Response status code does not indicate success: 400 (Snupkg upload failed. Please use latest NuGet clients (v 4.9.0 or above) and a V3 endpoint to push Symbol packages. For details, refer https://docs.microsoft.com/nuget/create-packages/symbol-packages-snupkg).
```

### DIFF
--- a/.pipelines/cs-build-steps.yaml
+++ b/.pipelines/cs-build-steps.yaml
@@ -12,7 +12,7 @@ steps:
     performMultiLevelLookup: true
     useGlobalJson: true
 
-- task: NuGetAuthenticate@0
+- task: NuGetAuthenticate@1
   displayName: 'NuGet Authenticate'
 
 - task: DotNetCoreCLI@2
--- a/.pipelines/cs-ci-build.yaml
+++ b/.pipelines/cs-ci-build.yaml
@@ -26,6 +26,10 @@ steps:
   inputs:
     signType: $(signType)
 
+- task: NuGetToolInstaller@1
+  inputs:
+    versionSpec: '6.4.0'
+
 # To archive our debug symbols with symweb, we need to convert the portable .pdb files that we build to windows .pdb files first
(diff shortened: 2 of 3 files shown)


.pipelines/cs-build-steps.yaml
```diff
@@ -12,7 +12,7 @@ steps:
     performMultiLevelLookup: true
     useGlobalJson: true
 
-- task: NuGetAuthenticate@0
+- task: NuGetAuthenticate@1
```
taiyosogawa, line 15: I noticed that the @0 version of this task is deprecated: https://learn.microsoft.com/en-us/azure/devops/pipelines/tasks/reference/nuget-authenticate-v0?view=azure-pipelines

.pipelines/cs-ci-build.yaml
```diff
@@ -26,6 +26,10 @@ steps:
   inputs:
     signType: $(signType)
 
+- task: NuGetToolInstaller@1
+  inputs:
+    versionSpec: '6.4.0'
```
jasongin, line 31: `'6.4.x'` is probably better, that way it can pick up patch updates which may include security fixes.

jasongin (approved):

.pipelines/cs-ci-build.yaml
taiyosogawa: Good point - thanks!

PR #193, opened 2023-02-07
Fix rust agent header and empty query string

taiyosogawa:
Cleans up the Rust http_client

### Changes proposed: 
- Ensure that the rust SDK version is included in the user agent header
- Fix a rust bug that appended `?` to URLs, even when there is no query string

### DIFF
--- a/rs/src/management/http_client.rs
+++ b/rs/src/management/http_client.rs
@@ -32,6 +32,8 @@ const TUNNELS_API_PATH: &str = "/api/v1/tunnels";
 const ENDPOINTS_API_SUB_PATH: &str = "endpoints";
 const PORTS_API_SUB_PATH: &str = "ports";
 const CHECK_TUNNEL_NAME_SUB_PATH: &str = "/checkNameAvailability";
+const PKG_VERSION: Option<&str> = option_env!("CARGO_PKG_VERSION");
+
 impl TunnelManagementClient {
     /// Returns a builder that creates a new client, starting with the current
     /// client's options.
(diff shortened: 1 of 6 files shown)


rs/src/management/http_client.rs
```diff
@@ -518,6 +499,33 @@ impl From<TunnelClientBuilder> for TunnelManagementClient {
     }
 }
 
+fn add_query(mut url: Url, tunnel_opts: &TunnelRequestOptions) -> Url
+{
```
connor4312, line 503: Please make sure to rust `rustfmt` for formatting. Turning on "format on save" with the rust analyzer extension will do this for you automatically, or you can manually run the "Format Document" command

rs/src/management/http_client.rs
```diff
@@ -472,12 +451,14 @@ pub struct TunnelClientBuilder {
 /// Creates a new tunnel client builder. You can set options, then use `into()`
 /// to get the client instance (or cast automatically).
 pub fn new_tunnel_management(user_agent: &str) -> TunnelClientBuilder {
+    let pkg_version = PKG_VERSION.unwrap_or("unknown");
+    let full_user_agent = format!("{}{}{}", user_agent, " Visual-Studio-Tunnel-Service-Rust-SDK/", pkg_version);
```
connor4312, line 455: Probably want to put a space or some kind of separator after the consumer-provided user agent.

rs/src/management/http_client.rs
```diff
@@ -518,6 +499,33 @@ impl From<TunnelClientBuilder> for TunnelManagementClient {
...
+        url.query_pairs_mut().append_pair("forceRename", "true");
+    }
+    if !tunnel_opts.tags.is_empty() {
+        url.query_pairs_mut().append_pair("tags", &tunnel_opts.tags.join(","));
+        if tunnel_opts.require_all_tags {
+            url.query_pairs_mut().append_pair("allTags", "true");
+        }
+    }
+    if tunnel_opts.limit > 0 {
+        url.query_pairs_mut().append_pair("limit", &tunnel_opts.limit.to_string());
```
connor4312, line 523: If this function took a `&mut Url`, you could keep the query pairs reference around as the original code did without having to return the Url. Not a big deal just less verbosity.

rs/src/management/http_client.rs
taiyosogawa: Good call! I ran the formatter and added a note to CONTRIBUTING.md.

rs/src/management/http_client.rs
taiyosogawa: Thanks for the tip - this is my first rust change and the feedback is appreciated. Done!

connor4312 (approved):

PR #194, opened 2023-02-08
Negotiate relay connection protocol version

jasongin:
This change adds support for negotiating the relay connection protocol version via the websocket subprotocol. Actual implementation of the v2 protocol is not included; that will come later.

 - Update `TunnelRelayStreamFactory` to support _requesting_ multiple protocols and _returning_ the single negotiated protocol selected by the service.
 - Add a public `ConnectionProtocol` property to tunnel host and client. (This property will primarily be used by the relay service to follow a different code path depending on the protocol version.)
 - Log the negotiated connection protocol in the verbose log.

### DIFF
--- a/cs/src/Connections/IRelayClient.cs
+++ b/cs/src/Connections/IRelayClient.cs
@@ -1,4 +1,4 @@
-﻿// <copyright file="IRelayClient.cs" company="Microsoft">
+// <copyright file="IRelayClient.cs" company="Microsoft">
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license.
 // </copyright>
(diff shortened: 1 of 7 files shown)


IlyaBiryukov:
You'd need to fix `MockTunnelRelayStreamFactory`

IlyaBiryukov (dismissed):

jasongin:
> You'd need to fix MockTunnelRelayStreamFactory

Yes, I'm preparing a set of related changes to support protocol version negotiation in the service.

jasongin:
Oh I see, there is a `MockTunnelRelayStreamFactory` also in the SDK test code.

IlyaBiryukov (approved):

PR #200, opened 2023-02-23
Add inspection uri to port contract

jfullerton44:
Fixes https://github.com/microsoft/basis-planning/issues/715

### Changes proposed: 
- Adds list of inspection URIs to tunnel port contract

### Other Tasks:

### DIFF
--- a/cs/src/Contracts/TunnelPort.cs
+++ b/cs/src/Contracts/TunnelPort.cs
@@ -147,8 +147,8 @@ public class TunnelPort
     public string[]? PortForwardingUris { get; set; }
 
     /// <summary>
-    /// Gets or sets inspection URIs.
-    /// If set, it's a list of absolute URIs where the port's traffic can be inspected.
+    /// Gets or sets inspection URI.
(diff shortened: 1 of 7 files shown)


cs/src/Contracts/TunnelPort.cs
```diff
@@ -145,4 +145,10 @@ public TunnelPort()
     /// If set, it's a list of absolute URIs where the port can be accessed with web forwarding.
     /// </summary>
     public string[]? PortForwardingUris { get; set; }
+
+    /// <summary>
+    /// Gets or sets inspection URIs.
+    /// If set, it's a list of absolute URIs where the port's traffic can be inspected.
+    /// </summary>
+    public string[]? InspectionUris { get; set; }
```
jasongin, line 153: Why is it more than one?

jasongin (approved):

cs/src/Contracts/TunnelPort.cs
jfullerton44: Good point, should just be one

jasongin (approved):

PR #204, opened 2023-03-10
Add SDK host API for accepting channel connections directly as streams instead of via TCP ports 

jramsay:
Fixes: Planning item #677

### Changes proposed: 
- Add AcceptForwardedPortConnections
- Add ForwardConnectionsToLocalPorts

### Other Tasks:


cs/src/Connections/ITunnelHost.cs
```diff
@@ -67,4 +73,12 @@ public interface ITunnelHost : IAsyncDisposable
     /// Connection status changed event.
     /// </summary>
     event EventHandler<ConnectionStatusChangedEventArgs>? ConnectionStatusChanged;
+
+    /// <summary>
+    /// An event which fires when a connection is made to the forwarded port.
+    /// </summary>
+    /// <remarks>
+    /// Set forwardConnectionsToLocalPorts to false if a local TCP socket should not be created for the connection stream.
```
jasongin, line 81: Clarify in comments that this event is only raised in that case.

cs/src/Connections/ITunnelHost.cs
```diff
@@ -28,6 +29,11 @@ public interface ITunnelHost : IAsyncDisposable
     /// </summary>
     Exception? DisconnectException { get; }
 
+    /// <summary>
+    /// A value indicating whether the port-forwarding service forwards connections to local TCP sockets.
```
jasongin, line 33: Mention the default value (true) in the doc-comments.

cs/src/Connections/TunnelRelayTunnelHost.cs
```diff
@@ -489,14 +490,17 @@ private void OnSshChannelOpening(object? sender, SshChannelOpeningEventArgs e)
                     portForwardRequest.Port);
                 e.FailureReason = SshChannelOpenFailureReason.AdministrativelyProhibited;
             }
+        } else if (portForwardRequest.ChannelType == "forwarded-tcpip")
```
jasongin, line 493: nit: `else` goes on a new line in C#.

jasongin (approved):

PR #205, opened 2023-03-14
Connection protocol v2 SDK implementation (C#)

jasongin:
This change implements the host and client SDK parts of the V2 relay connection protocol, for C# only. Ports to other SDK languages will come in separate PR(s).

For design & background info, refer to the [ADR](https://github.com/microsoft/basis-planning/blob/main/ADRs/0029-tunnel-connection-protocol.md).

Host and client SDKs have simultaneous support for V1 and V2 protocols; they prefer V2 but allow the service to choose V1 via the websocket subprotocol negotiation. So the V2 protocol is not used unless the service enables it, and that is currently conditional on a feature flag.

SDK client changes:
 - The client configures the SSH session to support negotiation of no encryption. Since that session is not E2E encrypted, it's more efficient to skip it. (V2 E2E encryption will be added back in in a future PR.)
 - The client sends its access token whenever it opens a new connection, using an extension of the `PortForwardChannelOpenMessage` class.

SDK host changes:
 - The host configures its SSH session with optional encryption, rather than using a SSH `MultiChannelStream`. That allows for more flexibility in dealing with session requests and port-forwarding.
 - The host sends its access token whenever it forwards a port, using an extension of the `PortForwardRequestMessage` class.
 - The host handles incoming client channels differently for V2 vs V1 because there's no longer an embedded session for each client. (Refer to the ADR.)

Additionally, the `TunnelAccessTokenProperties` now supports multiple port number claims in the token, that is necessary for scenarios when the client has limited access to only some port(s) but no tunnel-level access.

### DIFF
--- a/cs/src/Connections/TunnelClient.cs
+++ b/cs/src/Connections/TunnelClient.cs
@@ -44,6 +44,9 @@ public abstract class TunnelClient : TunnelConnection, ITunnelClient
     public ForwardedPortsCollection? ForwardedPorts =>
         SshPortForwardingService?.RemoteForwardedPorts;
 
+    /// <inheritdoc/>
+    public string? ConnectionProtocol { get; protected set; }
+
     /// <summary>
     /// Session used to connect to host
     /// </summary>
@@ -174,6 +177,18 @@ public abstract class TunnelClient : TunnelConnection, ITunnelClient
         // Enable reconnect only if connector is set as reconnect depends on it.
         var clientConfig = new SshSessionConfiguration(enableReconnect: this.connector != null);
 
+        if (ConnectionProtocol == TunnelRelayTunnelClient.WebSocketSubProtocolV2)
+        {
+            // Configure optional encryption, including "none" as an enabled and preferred kex algorithm,
+            // because encryption of the outer SSH session is optional since it is already over a TLS websocket.
+            clientConfig.KeyExchangeAlgorithms.Clear();
+            clientConfig.KeyExchangeAlgorithms.Add(SshAlgorithms.KeyExchange.None);
+            clientConfig.KeyExchangeAlgorithms.Add(SshAlgorithms.KeyExchange.EcdhNistp384);
+            clientConfig.KeyExchangeAlgorithms.Add(SshAlgorithms.KeyExchange.EcdhNistp256);
+            clientConfig.KeyExchangeAlgorithms.Add(SshAlgorithms.KeyExchange.DHGroup16Sha512);
+            clientConfig.KeyExchangeAlgorithms.Add(SshAlgorithms.KeyExchange.DHGroup14Sha256);
+        }
+
         // Enable port-forwarding via the SSH protocol.
         clientConfig.AddService(typeof(PortForwardingService));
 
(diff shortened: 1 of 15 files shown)


cs/src/Connections/TunnelClient.cs
```diff
@@ -161,10 +176,16 @@ protected async Task StartSshSessionAsync(Stream stream, CancellationToken cance
...
 
-        // For now, the client is allowed to skip SSH authentication;
-        // they must have a valid tunnel access token already to get this far.
-        var clientCredentials = new SshClientCredentials("tunnel", password: null);
-        await this.SshSession.AuthenticateAsync(clientCredentials);
+        // SSH authentication is required in V1 protocol, optional in V2 depending on whether the
+        // session enabled key exchange (as indicated by having a session ID or not). In either case
+        // a password is not required. Strong authentication was already handled by the relay
+        // service via the tunnel access token used for the websocket connection.
+        if (this.SshSession.SessionId != null)
```
IlyaBiryukov, line 183: Should we check whether `ConnectionProtocol` is v2 instead of `SessionId` being null to be consistent?

cs/src/Connections/TunnelConnection.cs
```diff
@@ -364,4 +368,75 @@ internal void OnRetrying(RetryingTunnelConnectionEventArgs e)
...
+        => Task.FromResult(new PortForwardSuccessMessage()); // Success messages are not extended.
+
+    Task<PortForwardChannelOpenMessage> IPortForwardMessageFactory.CreateChannelOpenMessageAsync(int port)
+        => Task.FromResult<PortForwardChannelOpenMessage>(
+            new PortRelayConnectRequestMessage { AccessToken = this.accessToken });
+
+    /// <summary>
+    /// Extends port-forward request messagse to include an access token property.
+    /// </summary>
+    private class PortRelayRequestMessage : PortForwardRequestMessage
```
IlyaBiryukov, line 386: Should we have these custom message classes as public contracts to ensure that both the server side and the client side use the same ones?

cs/src/Connections/TunnelClient.cs
IlyaBiryukov: Never mind, the comment explains why it's `SessionId`

IlyaBiryukov (dismissed):

cs/src/Connections/TunnelRelayTunnelClient.cs
```diff
@@ -127,12 +125,21 @@ protected Task ConnectAsync(string clientRelayUri, string? accessToken, Cancella
     /// </summary>
     protected virtual async Task<Stream> CreateSessionStreamAsync(CancellationToken cancellation)
     {
+        var protocols = Environment.GetEnvironmentVariable("DEVTUNNELS_PROTOCOL_VERSION") switch
```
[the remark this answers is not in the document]
jasongin: It is not exactly duplicated because it selects client protocols vs host protocols, and those are different constants. I don't see any reasonable way to make it shared.

cs/src/Connections/TunnelConnection.cs
jasongin: Done.

IlyaBiryukov (dismissed):

IlyaBiryukov (approved):

PR #208, opened 2023-03-16
Migrate from tslint to eslint

jasongin:
Summary:
 - Migrate from `tslint` to `eslint` with `typescript-eslint` plugins.
 - Update to latest `typescript`, `prettier`, and other supporting packages.
 - Fix linting issues in TypeScript code and/or tweak `.eslintrc.js` so that linting passes.
 - Delete unreferenced internal SshChannelForwarder class.
 - Update TS build pipelines from Node.js 12 to 18.


ts/.npmrc
```diff
@@ -1,3 +1 @@
 registy=https://registry.npmjs.org/
-
-always-auth=true
```
jasongin, line 3: `always-auth` was originally added when this project was using an internal Azure DevOps registry. It is not needed now when using the public npmjs registry. I'm removing it because the auth was failing in online builds after updating to Node.js 18.

jasongin: The problem was references to the internal feed still in `package-lock.json`. I rebuilt `package-lock.json` using only the public feed.

IlyaBiryukov (approved):

PR #215, opened 2023-04-04
Enable optional E2E encryption with v2 protocol

jasongin:
- The `ForwardedPortConnectingEventArgs` has moved down into the SSH library, so it is deleted here.
 - Add `PortRelayConnectResponseMessage` to go along with the `PortRelayConnectRequestMessage`, and send/receive these messages on client and host. These extend the SSH channel-open messages to enable the tunnel client, host, and relay to negotiate whether E2E encryption is enabled for each connection (without adding any additional round trips).
 - When E2E encryption is enabled for a connection, set up an SSH `SecureStream` as a transform on the host and client connection stream.

The v2 connection protocol is C# only for now. Porting to other languages will come in separate PR(s).


cs/src/Connections/TunnelClient.cs
```diff
@@ -221,19 +225,62 @@ protected async Task StartSshSessionAsync(Stream stream, CancellationToken cance
...
+
+    /// <summary>
+    /// Invoked when a forwarded port is connecting.
+    /// </summary>
+    protected virtual void OnForwardedPortConnecting(
+        object? sender, ForwardedPortConnectingEventArgs e)
+    {
+        var channel = e.Stream.Channel;
+        var relayResponseMessage = channel.OpenConfirmationMessage
+            .ConvertTo<PortRelayConnectResponseMessage>();
```
IlyaBiryukov, line 255: How is this going to work during the transition period, when the open confirmation message is not yet actually `PortRelayConnectResponseMessage` (i.e. the service still sends out the old message) ? Will it cause binary protocol incompatibility?

jasongin: The `OnForwardedPortConnecting()` method is only hooked up for V2 protocol. So it's a breaking change for V2, but nobody is using that version yet.

I'll add more comments clarifying that.

cs/src/Connections/TunnelRelayTunnelHost.cs
```diff
@@ -295,16 +298,19 @@ private void HostSession_ChannelOpening(object? sender, SshChannelOpeningEventAr
             e.Channel.ChannelType == "forwarded-tcpip")
         {
             // V2 protocol.
+            var relayRequestMessage = e.Channel.OpenMessage
+                .ConvertTo<PortRelayConnectRequestMessage>();
```
IlyaBiryukov, line 302: Same question here. What if the service sends out the old message, will it cause any trouble?

cs/src/Connections/TunnelRelayTunnelHost.cs
```diff
@@ -345,6 +351,61 @@ void RemoveClientSessionTask(Task t)
...
+            {
+                try
+                {
+                    // Do not pass the cancellation token from the connecting event,
+                    // because the connection will outlive the event.
+                    await secureStream.ConnectAsync();
+                }
+                catch (Exception ex)
+                {
+                    channel.Trace.Error(
```
IlyaBiryukov, line 400: Let's catch any exceptions out of `channel.Trace.Error` to be safe in `async void` method.

cs/src/Connections/TunnelRelayTunnelHost.cs
jasongin: It's V2 protocol only.

Note the current V2 websocket subprotocol has the "-dev" suffix. When V2 is stable we'll drop the "-dev" suffix, so any non-updated clients will not connect with an incomplete implementation of V2 protocol.

IlyaBiryukov (dismissed):

cs/src/Connections/TunnelRelayTunnelHost.cs
jasongin: Added comments.

cs/src/Connections/TunnelRelayTunnelHost.cs
jasongin: Done.

IlyaBiryukov (approved):

PR #216, opened 2023-04-07
Fix TunnelConnection.GetFreshAccessToken 

IlyaBiryukov:
Fix for [#806](https://github.com/microsoft/basis-planning/issues/806)

When refreshing tunnel access token, `TunnelConnection.GetFreshAccessToken()` calls `tunnelManager` if it was set to fetch the current tunnel using `Tunnel` property. If `Tunnel` has a valid access token, then fetching succeeds but the fetched tunnel may not have the access token, and so a valid access token on `Tunnel` object may be ignored.

The fix is to use the access token from `Tunnel` if the fetched tunnel doesn't have one.

Add `TunnelExtensions` class with helper methods to fetch tunnel tokens from `Tunnel` object. 
Add unit tests for it.

### DIFF
--- a/cs/src/Connections/TunnelConnection.cs
+++ b/cs/src/Connections/TunnelConnection.cs
@@ -1,4 +1,4 @@
-﻿// <copyright file="TunnelConnection.cs" company="Microsoft">
+// <copyright file="TunnelConnection.cs" company="Microsoft">
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license.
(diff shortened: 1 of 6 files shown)


jasongin:
Should these changes be ported to other language SDKs?

jasongin (approved):

IlyaBiryukov:
The fix to the other SDKs (only ts currently) that refresh tunnel access token will come in a separate PR

PR #222, opened 2023-04-13
Send port-forward request messages concurrently

jasongin:
This is a small optimization that can make client connection setup a little faster, particularly when there are a lot of ports forwarded over a high-latency connection.

The change is only in the C# SDK, because the TS SSH lib doesn't support concurrent session requests. However it can receive and handle those concurrent requests (sequentially); I confirmed there is no compatibility here.

(The C# SDK didn't support concurrent requests either at the time this code was originally written. That support was added more recently.)

### DIFF
--- a/cs/src/Connections/TunnelRelayTunnelHost.cs
+++ b/cs/src/Connections/TunnelRelayTunnelHost.cs
@@ -558,25 +558,20 @@ public class TunnelRelayTunnelHost : TunnelHost, IRelayClient
     private async Task StartForwardingExistingPortsAsync(
         SshSession session, bool removeUnusedPorts = false)
     {
+        // Send port-forward request messages concurrently. The client may still handle the
+        // requests sequentially but at least there is no network round-trip between them.
+        var forwardTasks = new List<Task>();
+
         var tunnelPorts = Tunnel!.Ports ?? Enumerable.Empty<TunnelPort>();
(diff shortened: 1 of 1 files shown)


cs/src/Connections/TunnelRelayTunnelHost.cs
```diff
@@ -558,26 +558,21 @@ private void OnSshClientAuthenticating(object? sender, SshAuthenticatingEventArg
...
-            {
-                Trace.TraceEvent(
-                    TraceEventType.Error,
-                    0,
-                    "Error forwarding port {0} to client: {1}",
-                    port.PortNumber,
-                    exception.Message);
-            }
+            // ForwardPortAsync() catches and logs most exceptions that might normally occur.
+            forwardTasks.Add(ForwardPortAsync(pfs, port, CancellationToken.None));
```
IlyaBiryukov, line 571: Before the change the exceptions from `ForwardPortAsync` were traced and ignored. Now they'd pop up to the caller. Is this intended?

jasongin: Yes - see the comment I added on line 570. The catch here was redundant, unless some unexpected kind of exception occurred, and in that case a crash is probably preferred over ignoring it.

IlyaBiryukov (approved):

PR #226, opened 2023-04-25
Update user-agent in SDKs to use Dev-Tunnels-Service

derekbekoe:


jfullerton44 (changes requested):
Can you update https://github.com/microsoft/dev-tunnels/blob/203926bff66d0e0852b978980eebdeed89616cc8/go/tunnels/tunnels.go#L13

jfullerton44 (approved):

PR #227, opened 2023-04-25
Go SDK: Ensure tunnels.go is updated when there's a change to the go sdk folder

derekbekoe:
Helps to catch when PackageVersion is not updated when the PR includes a Go SDK change.

### Changes proposed: 
- Add GitHub action check.
  - First checks that the `go` directory and `go/tunnels/tunnels.go` file exists. This was if they get moved, the action will error out and catch it.
  - If Go SDK folder is updated, it ensures `go/tunnels/tunnels.go` was also updated since that contains `PackageVersion`.

### Other Tasks:
- [n/a] If you updated the Go SDK did you update the PackageVersion in tunnels.go

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -33,7 +33,7 @@ jobs:
     - name: Test
       run: cd go/tunnels && go test -short -v ./...
 
-  check-package-version-update:
+  check-version-update:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
@@ -51,5 +51,7 @@ jobs:
           exit 0
         else
           echo "Error: An error occurred. Has "PackageVersion" in $FILE_WITH_PKG_VER been updated?"
+          git diff --name-only $BASE_SHA HEAD
+          git diff $BASE_SHA HEAD -- $FILE_WITH_PKG_VER
           exit 1
         fi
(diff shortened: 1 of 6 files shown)


.github/workflows/go.yml
```diff
@@ -32,3 +32,28 @@ jobs:
...
+        FILE_WITH_PKG_VER=go/tunnels/tunnels.go
+        test -d $SDK_DIR && test -f $FILE_WITH_PKG_VER
+        BASE_SHA=$(jq -r '.pull_request.base.sha' "$GITHUB_EVENT_PATH")
+        # If go sdk dir updated, package version file updated and "PackageVersion" line updated, success; else error.
+        if git diff --name-only $BASE_SHA HEAD | grep -q "^$SDK_DIR"; then
+          if git diff --name-only $BASE_SHA HEAD | grep -q "^$FILE_WITH_PKG_VER" && git diff $BASE_SHA HEAD -- $FILE_WITH_PKG_VER | grep -q "PackageVersion"; then
+            echo "Success: Package version was updated."
+            exit 0
+          else
+            echo "Error: An error occurred. Has "PackageVersion" in $FILE_WITH_PKG_VER been updated?"
```
IlyaBiryukov, line 54: Should the error go to stderr stream?

IlyaBiryukov (approved):

.github/workflows/go.yml
derekbekoe: updated

jasongin (approved):

PR #231, opened 2023-05-03
Port connection protocol v2 to TS SDK

jasongin:
This is a port of changes from the following C# PRs to the TS SDK:
 - #205
 - #215
 - #223


ts/src/connections/relayTunnelConnector.ts
```diff
@@ -123,8 +123,12 @@ export class RelayTunnelConnector implements TunnelConnector {
             disconnectReason = SshDisconnectReason.connectionLost;
             error = undefined;
             try {
-                stream = await this.tunnelSession.createSessionStream(cancellation);
-                await this.tunnelSession.configureSession(stream, isReconnect, cancellation);
+                const streamAndProtocol = await this.tunnelSession.createSessionStream(
```
IlyaBiryukov, line 126: Nit: can we decompose to stream and protocol in place?

ts/src/connections/tunnelConnectionSession.ts
```diff
@@ -298,4 +323,25 @@ export class TunnelConnectionSession extends TunnelConnectionBase implements Tun
             return this.accessToken;
         }
     }
+
+    /** @internal */
+    public createRequestMessageAsync(port: number): Promise<PortForwardRequestMessage> {
```
IlyaBiryukov, line 328: Will this internal status be reflected in d.ts? If we wanted to hide these from the public surface area, we could create a separate internal class that would implement these methods.

ts/src/connections/tunnelHostBase.ts
```diff
@@ -53,14 +55,20 @@ export class TunnelHostBase
      */
     public hostPrivateKeyPromise?: Promise<KeyPair>;
 
+    /**
```
IlyaBiryukov, line 58: Should we move `connectionProtocol` property to the base class? This way both host and client can use it.

ts/src/connections/tunnelRelayStreamFactory.ts
```diff
@@ -11,14 +11,14 @@ export interface TunnelRelayStreamFactory {
...
      * @param clientConfig Client config for websocket.
      */
     createRelayStream(
         relayUri: string,
-        connectionType: string,
+        protocols: string[],
         accessToken?: string,
         clientConfig?: IClientConfig,
-    ): Promise<Stream>;
+    ): Promise<{ stream: Stream, protocol: string }>;
```
IlyaBiryukov, line 23: How about declaring an interface for `{ stream: Stream, protocol: string }` ? It would be easier for implementers.

ts/src/connections/tunnelRelayTunnelHost.ts
```diff
@@ -39,17 +46,32 @@ import { SshHelpers } from './sshHelpers';
 import { MultiModeTunnelHost } from './multiModeTunnelHost';
 import { TunnelHostBase } from './tunnelHostBase';
 import { tunnelRelaySessionClass } from './tunnelRelaySessionClass';
+import { SessionPortKey } from './sessionPortKey';
+import { PortRelayConnectRequestMessage } from './messages/portRelayConnectRequestMessage';
+import { PortRelayConnectResponseMessage } from './messages/portRelayConnectResponseMessage';
 
 const webSocketSubProtocol = 'tunnel-relay-host';
```
IlyaBiryukov, line 53: Aren't these sub protocol constants already defined in `tunnelClientBase` ?

ts/src/connections/tunnelClientBase.ts
```diff
@@ -29,6 +33,10 @@ import { Emitter } from 'vscode-jsonrpc';
 import { TunnelConnectionSession } from './tunnelConnectionSession';
 import { TunnelManagementClient } from '@microsoft/dev-tunnels-management';
 import { tunnelSshSessionClass } from './tunnelSshSessionClass';
+import { PortRelayConnectResponseMessage } from './messages/portRelayConnectResponseMessage';
+
+export const webSocketSubProtocol = 'tunnel-relay-client';
```
IlyaBiryukov, line 38: Nit: can we use `clientWebSocketSubProtocol` and `hostWebSocketSubProtocol` constant names? Otherwise they may get confused which is which.

ts/src/connections/tunnelRelayTunnelHost.ts
IlyaBiryukov, line 53: Maybe move these to `tunnelHostBase`? The client's ones are in `tunnelClientBase`

ts/src/connections/tunnelRelayTunnelHost.ts
```diff
@@ -39,17 +46,32 @@ import { SshHelpers } from './sshHelpers';
...
 /**
  * Tunnel host implementation that uses data-plane relay
  *  to accept client connections.
  */
 export class TunnelRelayTunnelHost extends tunnelRelaySessionClass(
     TunnelHostBase,
-    webSocketSubProtocol,
+    connectionProtocols,
 ) {
+    public static readonly webSocketSubProtocol = webSocketSubProtocol;
```
IlyaBiryukov, line 72: Can these two statics go to `tunnelRelaySessionClass`? This way they'll be shared with `TunnelRelayTunnelClient`.

jasongin: The client and host protocol names are not the same, so they can't be shared in that way.

ts/src/connections/tunnelConnectionSession.ts
jasongin: The `@internal` tag causes them to be omitted from the `.d.ts`.

It could be possible to refactor to a helper class, but `@internal` works fine.

ts/src/connections/relayTunnelConnector.ts
jasongin: The `stream` is already defined, while `protocol` is not, so I found a destructuring assignment to be just as awkward as this.

ts/src/connections/tunnelRelayTunnelHost.ts
jasongin: The public properties are on the subclasses in both cases.

In the client case, I had to define the constants in the base source file (then import them in the subclass source file) because they are used in some implementation code in the base class.

I'm a little concerned that we might have more levels of inheritance than necessary:
`TunnelRelayTunnelClient`->`TunnelClientBase`->TunnelConnectionSession`->`TunnelConnectionBase` plus mix-ins `tunnelRelaySessionClass` and `tunnelSshSessionClass`. It has made it hard for me to find things or know where to put new code.

Some of that was supposed to allow for other connection implementations (such as LS or local network) but now I don't know if we will ever need that flexibility.

ts/src/connections/tunnelRelayStreamFactory.ts
jasongin: It is a `(Stream, string)` tuple in the C# SDK. I don't think this is difficult for implementors, and I prefer to keep them in sync. In fact this kind of pattern is much more common in TypeScript.

ts/src/connections/tunnelClientBase.ts
jasongin: The _public_ constants are distinct because they are properties of different classes. These constants are private; they are exported only for use by the base class, but are not public because they are not exported by `index.ts`.

ts/src/connections/tunnelHostBase.ts
jasongin: Done.

ts/src/connections/tunnelRelayTunnelHost.ts
IlyaBiryukov: If we don't expect to subclass any of the classes in the inheritance chain, I'm all for simplifying it. Maybe not part of this PR though.

IlyaBiryukov (approved):

PR #241, opened 2023-05-24
Create rust.yml

usanth:
Fixes #

### Changes proposed: 
Adding a github action to build and check rust files

### Testing:
I added this and then this check ran as expected when this PR was getting built

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -1,0 +1,23 @@
+name: Rust
+
+on:
+  push:
+    branches: [ "main" ]
+  pull_request:
+    branches: [ "main" ]
+
+env:
+  CARGO_TERM_COLOR: always
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v3
+    - name: Build
+      run: cargo build --verbose --manifest-path=rs/Cargo.toml
+    - name: Run tests
+      run: cargo test --verbose --manifest-path=rs/Cargo.toml
+   
--- a/rs/src/contracts/named_rate_status.rs
+++ b/rs/src/contracts/named_rate_status.rs
@@ -13,5 +13,5 @@ pub struct NamedRateStatus {
     pub base: RateStatus,
 
     // The name of the rate status.
-    pub name: Option<String>,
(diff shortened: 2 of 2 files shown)


.github/workflows/rust.yml
```diff
@@ -0,0 +1,23 @@
...
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v3
+    - name: Build
+      run: cargo build --verbose --manifest-path=rs/Cargo.toml
```
IlyaBiryukov, line 20: Did you run it with errors in .rs files? Does it fail the PR in this case?

IlyaBiryukov (dismissed):

.github/workflows/rust.yml
usanth: Let me try that now and if so then I will complete the PR

usanth: When I added a syntax error in a rust file the action broke: https://github.com/microsoft/dev-tunnels/actions/runs/5073570643/jobs/9112748431?pr=241 I will revert this error now since it works

IlyaBiryukov, line 20: Can this build execute any arbitrary code from the files being checked in? E.g. if it was package.json - it could run scripts, and somebody could add malicious code and try to run it in the PR workflow. 

If it can, please configure the workflow to only run in PRs from trusted users.

IlyaBiryukov (approved):

.github/workflows/rust.yml
jasongin: It's a setting on the repo, not on individual workflows.

IDK if you can see it in the repo settings, but here is what it looks like for me:

image

jasongin: So if any external user submits a PR there will be a button on it that allows only team members to approve the workflows for the PR. It is up to us to review the changes enough to verify there are is no malicious code before approving.

usanth: This is great, thanks for pointing this out Jason! I think in that case this doesn't have to do with this specific PR and instead is set for the entire repo, so I will go ahead and resolve this comment

PR #248, opened 2023-06-05
Dev/jarober/update cluster id regex

JacobRoberts:
### Changes proposed: 
- Update cluster ID regex to accurately represent dev tunnels cluster IDs

### Other Tasks:

### DIFF
--- a/cs/src/Contracts/TunnelConstraints.cs
+++ b/cs/src/Contracts/TunnelConstraints.cs
@@ -137,7 +137,7 @@ public static class TunnelConstraints
     /// Cluster IDs are alphanumeric; hyphens are not permitted.
     /// </remarks>
     /// <seealso cref="Tunnel.ClusterId"/>
-    public const string ClusterIdPattern = "[a-z][a-z0-9]{2,11}";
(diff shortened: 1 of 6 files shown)


cs/src/Contracts/TunnelConstraints.cs
```diff
@@ -137,7 +137,7 @@ public static class TunnelConstraints
     /// Cluster IDs are alphanumeric; hyphens are not permitted.
     /// </remarks>
     /// <seealso cref="Tunnel.ClusterId"/>
-    public const string ClusterIdPattern = "[a-z][a-z0-9]{2,11}";
+    public const string ClusterIdPattern = "^(([a-z]{3,4}[0-9]{1,3})|asse|aue|brs|euw|use)$";
```
IlyaBiryukov, line 140: Should we include "localhost1" and "localhost2" cluster IDs used by testing?

JacobRoberts: Discussed with Ilya, the solution for us is to use valid cluster IDs instead of `localhost1` and `localhost2` in our testing environments.

PR #254, opened 2023-06-19
Dev/utsasanthosh/proxy connection ts

usanth:
Fixes #https://github.com/microsoft/basis-planning/issues/758

### Changes proposed: 
- If a http proxy is provided, that will be used when creating the web socket connection to the relay service
- The proxy will be provided by getting passed into the `TunnelManagementHttpClient` at initialization, and then will be used when the `TunnelManagementHttpClient` is passed into the `TunnelRelayTunnelClient` to make the connection
- This will unblock VSCode Remote Tunnels

### Other Tasks:
- Tested this by creating a sample project

### DIFF
--- a/ts/src/connections/sshHelpers.ts
+++ b/ts/src/connections/sshHelpers.ts
@@ -8,7 +8,6 @@ import {
     connection as WebSocketConnection,
     IClientConfig,
 } from 'websocket';
-import * as http from 'http';
  
 
 declare module 'websocket' {
--- a/ts/src/connections/sshHelpers.ts
+++ b/ts/src/connections/sshHelpers.ts
@@ -141,8 +141,6 @@ export class SshHelpers {
                 resolve(new ssh.WebSocketStream(new WebsocketStreamAdapter(connection)));
             });
 
-            console.log("clientConfig = ", clientConfig?.tlsOptions?.agent);
-
             // If the server responds but doesn't properly upgrade the connection to web socket, WebSocketClient fires 'httpResponse' event.
             // TODO: Return ProblemDetails from TunnelRelay service
             client.on('httpResponse', ({ statusCode, statusMessage }) => {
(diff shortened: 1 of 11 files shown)


ts/src/management/tunnelManagementHttpClient.ts
```diff
@@ -903,4 +903,4 @@ export class TunnelManagementHttpClient implements TunnelManagementClient {
 
         return content;
     }
```
IlyaBiryukov, line 905: Is the tunnel management client already capable of using `http.Agent` ?

ts/src/connections/tunnelConnector.ts
```diff
@@ -12,5 +13,5 @@ export interface TunnelConnector {
      * @param isReconnect A value indicating if this is a reconnect (true) or regular connect (false).
      * @param cancellation Cancellation token.
      */
-    connectSession(isReconnect: boolean, cancellation: CancellationToken): Promise<void>;
+    connectSession(isReconnect: boolean, cancellation: CancellationToken, httpAgent?: http.Agent): Promise<void>;
```
jasongin, line 16: By convention, `cancellation` should be the final parameter. So insert `httpAgent` before it.

ts/src/management/tunnelManagementHttpClient.ts
usanth: Could you elaborate on this? Initially without these changes, the `TunnelManagementHttpClient` had the `http.Agent` as an optional parameter. The issue is that was never passed in as part of the web socket connection which is what this PR aims to solve

IlyaBiryukov: Thanks! that's what I was asking about!

ts/src/connections/sshHelpers.ts
```diff
@@ -8,6 +8,8 @@ import {
     connection as WebSocketConnection,
     IClientConfig,
 } from 'websocket';
+import * as http from 'http';
```
IlyaBiryukov, line 11: Unused?

IlyaBiryukov (approved):

ts/src/connections/tunnelConnector.ts
jasongin: Same in other places.

jasongin (approved):

ts/src/connections/tunnelConnector.ts
usanth: When I attempt to do that it says that since `httpAgent` is an optional parameter it cannot be followed by required parameters which is why I put it after the cancellation token. The reason I didn't make it a required parameter was because I thought that there would be instances where you wouldn't have to pass the proxy in for connection.

PR #261, opened 2023-06-20
[Contracts] Allow empty string in name & domain validation.

jramsay:
Fixes # https://github.com/microsoft/basis-planning/issues/925

### Changes proposed: 
- Allow empty string in tunnel name & domain pattern
- Tested consuming the change in the CLI

### Other Tasks:

### DIFF
--- a/cs/src/Contracts/TunnelConstraints.cs
+++ b/cs/src/Contracts/TunnelConstraints.cs
@@ -183,7 +183,7 @@ public static class TunnelConstraints
     /// allows an empty string because tunnels may be unnamed.
     /// </remarks>
     /// <seealso cref="Tunnel.Name"/>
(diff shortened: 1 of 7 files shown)


IlyaBiryukov:
What is the scenario when tunnel name or domain is empty string?

jramsay:
@IlyaBiryukov: it is when you're updating the tunnel name or domain. You should also be able to set it back to the default - no name or domain. Our service has logic in it that when it receives an empty string it sets it back to null. This was brought up in a PR comment by @jasongin in another PR. I filed a separate item to resolve this as it involved contract changes to the validation patterns.

jfullerton44 (approved):

jramsay:
Note: I didn't update IsValidTunnelName(string tunnelName)

This has a string.IsNullOrEmpty(tunnelName) check that will fail. It doesn't get used in the update scenario that we wish to support, and it used in other places (ex: tunnel plan names validation) where an empty string might not be desirable.

PR #264, opened 2023-06-28
C# and node client can refresh host public keys on connection.

IlyaBiryukov:
Fix for [#979](https://github.com/microsoft/basis-planning/issues/979).

When the client fails to authenticate the host SSH session, it must refresh the tunnel from the service and get the new endpoint public key if possible, then retry.

* Add new connection status when the client refreshes the tunnel to get the new host public keys.
* Refactor various places where the tunnel object properties are used for tunnel host and client.
* Make authentication callback in the client handle host public key mismatch by fetching the tunnel and retrying.
* Add unit tests.
* Change how `httpAgent` is passed to ts SDK clients.
* Fix ts mock tunnel management client getTunnel() not returning the right tunnel.
* Fix ts SDK unit tests incorrectly using v2 connection protocol.

### DIFF
--- a/cs/src/Connections/ConnectionStatus.cs
+++ b/cs/src/Connections/ConnectionStatus.cs
@@ -1,4 +1,4 @@
-﻿// <copyright file="ConnectionStatus.cs" company="Microsoft">
+// <copyright file="ConnectionStatus.cs" company="Microsoft">
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license.
 // </copyright>
@@ -34,4 +34,9 @@ public enum ConnectionStatus
     /// Disconnected from the service and could not reconnect either due to disposal, service down, tunnel deleted, or token expiration. This is the final status.
     /// </summary>
     Disconnected,
+
+    /// <summary>
+    /// Refreshing tunnel host public key.
+    /// </summary>
+    RefreshingTunnelHostPublicKey,
 }
--- a/cs/src/Connections/TunnelConnection.cs
+++ b/cs/src/Connections/TunnelConnection.cs
@@ -25,6 +25,7 @@ public abstract class TunnelConnection : IAsyncDisposable, IPortForwardMessageFa
     private readonly CancellationTokenSource disposeCts = new();
     private Task? reconnectTask;
     private ConnectionStatus connectionStatus;
+    private Tunnel? tunnel;
 
     /// <summary>
     /// Creates a new instance of the <see cref="TunnelConnection"/> class.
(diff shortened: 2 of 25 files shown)


usanth:
The `httpAgent` work looks good to me! I tested it with a proxy and everything works as expected. Waiting on others to review as well

cs/src/Connections/ConnectionStatus.cs
```diff
@@ -34,4 +34,9 @@ public enum ConnectionStatus
     /// Disconnected from the service and could not reconnect either due to disposal, service down, tunnel deleted, or token expiration. This is the final status.
     /// </summary>
     Disconnected,
+
+    /// <summary>
+    /// Refreshing tunnel host public key.
+    /// </summary>
+    RefreshingTunnelHostPublicKey,
```
jasongin, line 41: Do we need a separate status value for this? Should it just be part of `Connecting` status?

cs/src/Connections/TunnelConnection.cs
```diff
@@ -78,7 +79,32 @@ protected set
...
-    public Tunnel? Tunnel { get; private set; }
+    public Tunnel? Tunnel {
+        get => this.tunnel;
+        private set
+        {
+            if (value != this.tunnel)
+            {
+                // Get the tunnel access token from the new tunnel, or the original Tunnal object if the new tunnel doesn't have the token,
+                // which may happen when the tunnel was authenticated with a tunnel access token from Tunnel.AccessTokens.
+                // Add the tunnel access token to the new tunnel's AccessTokens if it is not there.
```
jasongin, line 90: I was planning to address this in the management APIs. See https://github.com/microsoft/basis-planning/issues/990

cs/src/Connections/TunnelConnection.cs
```diff
@@ -249,18 +283,22 @@ protected async Task<bool> RefreshTunnelAccessTokenAsync(CancellationToken cance
...
 
-            return false;
+            // Access token may be null if tunnel allows anonymous access.
+            if (this.accessToken != null)
+            {
+                TunnelAccessTokenProperties.ValidateTokenExpiration(this.accessToken);
+            }
+
+            Trace.TraceInformation(
+                "Refreshed tunnel access token. New token: {0}",
```
jasongin, line 298: Avoid logging the full token, especially in non-verbose logs. I see this line was moved from above, but still...

jasongin (approved):

cs/src/Connections/ConnectionStatus.cs
IlyaBiryukov: Since there is `RefreshingTunnelAccessToken` which is also part of (re-)connection flow, I though adding `RefreshingTunnelHostPublicKey` may help with transparency into the current client status, and with unit tests.

cs/src/Connections/TunnelConnection.cs
IlyaBiryukov: Logging verbose now.

cs/src/Connections/ConnectionStatus.cs
jasongin: > there is `RefreshingTunnelAccessToken`

I forgot we had that one already. I'm a little concerned that as we're likely to add/remove/reorder different (optional) sub-steps of connecting, maybe that detail shouldn't be exposed to the application. But maybe it's fine.

jasongin:
@IlyaBiryukov is this ready to merge? I'm working on a change (V2 reconnect) that touches some of the same areas of code.

jasongin (approved):

PR #268, opened 2023-07-06
change Client ssh info logs  to verbose.

setaskin:
Fixes #840

### Changes proposed: 
- This Pr  makes some ClientSSH logs verbose. So will be cleaner for non-verbose logs in CLI. 
- Session closed and cancelled information should start with Client SSH for consistency

### Other Tasks:

### DIFF
--- a/cs/src/Connections/TunnelRelayTunnelHost.cs
+++ b/cs/src/Connections/TunnelRelayTunnelHost.cs
@@ -162,7 +162,7 @@ public class TunnelRelayTunnelHost : TunnelHost, IRelayClient
         };
 
         ValidateAccessToken();
-        Trace.TraceInformation("Connecting to host tunnel relay {0}", this.relayUri!.AbsoluteUri);
+        Trace.Verbose("Connecting to host tunnel relay {0}", this.relayUri!.AbsoluteUri);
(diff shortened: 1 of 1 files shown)


cs/src/Connections/TunnelRelayTunnelHost.cs
```diff
@@ -495,11 +495,12 @@ void Session_Closed(object? sender, SshSessionClosedEventArgs e)
             // Reconnecting client session may cause the new session close with 'None' reason and null exception.
             if (cancellation.IsCancellationRequested)
             {
-                Trace.TraceInformation("Client ssh session cancelled.");
+                Trace.WithName("ClientSSH").Verbose("Session cancelled verbose.");
```
IlyaBiryukov, line 498: Please don't add "verbose" suffix to the trace message. It's already traced as verbose via `.Verbose()` method.

setaskin: Sorry Yes, it was a copy paste mistake. I will fix it

setaskin: fixed

IlyaBiryukov (approved):

PR #278, opened 2023-07-14
Reconnect V2 protocol

jasongin:
This implements reconnection for the tunnel protocol V2.

It depends on [the SecureStream reconnection PR in the SSH repo](https://github.com/microsoft/dev-tunnels-ssh/pull/58). Before merging this PR I'll update the SSH library version to pick up the SecureStream changes after that PR is merged. Builds of this PR will fail until then.

On the host side, reconnection is entirely handled by the `SecureStream` server; the only change is to keep track of the disconnected/reconnectable sessions for each tunnel host.

On the client side, the tunnel client keeps track of disconnected `SecureStream`s for each port, and then triggers reconnection of streams in response to `PortUpdated` events which occur after the outer (unencrypted) session connection is re-established.

Note tests for V2 connections and reconnection are in the service repo (not here) because the relay service implements an important part of the protocol. But I had to update some existing tests that were broken by the addition of a `Disconnected` status event before reconnecting.


ts/test/tunnels-test/tunnelHostAndClientTests.ts
```diff
@@ -146,11 +146,11 @@ export class TunnelHostAndClientTests {
             clientStreamFactory,
         );
 
-        assert.strictEqual(false, relayClient.isSshSessionActiveProperty);
+        assert.strictEqual(relayClient.isSshSessionActiveProperty, false);
```
jasongin, line 149: I fixed the assertion actual/expected parameter order in many places in this file.

ts/test/tunnels-test/tunnelHostAndClientTests.ts
```diff
@@ -464,24 +464,31 @@ export class TunnelHostAndClientTests {
 
     @test
     public async connectRelayClientAddPortInUse() {
```
jasongin, line 466: This test didn't work at all before; it was trying to use `socket.connect()` instead of `socket.listen()` to create a port conflict. It appeared to succeed only because it never awaited the bad code.

cs/src/Connections/TunnelClient.cs
```diff
@@ -254,6 +258,43 @@ private void ConfigurePortForwardingService()
...
+        if (!port.HasValue)
+        {
+            return;
+        }
+
+        var disconnectedStreamsCount = 0;
+        lock (this.disconnectedStreams)
+        {
+            // If there are disconnected streams for the port, re-connect them now.
+            if (this.disconnectedStreams.TryGetValue(port.Value, out var streams))
```
taiyosogawa, line 278: Are we supposed to remove the port from `disconnectedStreams` if reconnecting the streams is successful?

cs/src/Connections/TunnelClient.cs
```diff
@@ -254,6 +258,43 @@ private void ConfigurePortForwardingService()
...
+            }
+        }
+
+        for (int i = 0; i < disconnectedStreamsCount; i++)
+        {
+            Task.Run(async () =>
+            {
+                try
+                {
+                    await pfs.ConnectToForwardedPortAsync(port.Value, CancellationToken.None);
```
taiyosogawa, line 290: I'm not too familiar with how channels work - do they not have a name or ID? I'm just wondering why it's not necessary to pass the value of `i` or `streams[i]`?

cs/src/Connections/TunnelClient.cs
```diff
@@ -254,6 +258,43 @@ private void ConfigurePortForwardingService()
...
+                }
+                catch (Exception ex)
+                {
+                    this.Trace.Warning(
+                        $"Failed to reconnect to forwarded port {port}: {ex.Message}");
```
taiyosogawa, line 295: What would happen if 1 of 5 streams failed to reconnect? I'm wondering if we should just keep trying `pfs.ConnectToForwardedPortAsync()` until we either succeed or encounter X failures. (Perhaps it would be such a rare event for a partial failure that it's not worth it?)

cs/src/Connections/TunnelClient.cs
jasongin: That is done on line 339.

This method restores channels for the forwarded port, then the `OnForwardedPortConnectiong()` method below reconnects the `SecureStream` for each channel.

cs/src/Connections/TunnelClient.cs
jasongin: SSH channels have an integer ID, however we don't store a mapping from channel ID to `SecureStream` instance because the channel IDs may be different after the reconnection.

The SSH session inside `SecureStream` has a (cryptographically strong) ID that is used to match the streams when reconnecting, as commented on line 332.

cs/src/Connections/TunnelClient.cs
jasongin: If one of the streams fails to reconnect, then after a 60s timeout the application using the stream would eventually see it as a socket disconnection.

I don't think retrying here is likely to be helpful. If the connection to the forwarded port fails that means the host application is no longer listening on that port on the host. So, actually in that case we should immediately close any _disconnected_ streams for that port, so client applications don't have to wait for a timeout. I'll make that code change.

jasongin: Updated.

ts/src/connections/tunnelClientBase.ts
```diff
@@ -236,6 +250,36 @@ export class TunnelClientBase
...
+    private onForwardedPortAdded(pfs: PortForwardingService, e: ForwardedPortEventArgs) {
+        const port = e.port.remotePort;
+        if (typeof port !== 'number') {
+            return;
+        }
+
+        // If there are disconnected streams for the port, re-connect them now.
+        const disconnectedStreamsCount = this.disconnectedStreams.get(port)?.length ?? 0;
+        for (let i = 0; i < disconnectedStreamsCount; i++) {
+            pfs.connectToForwardedPort(port).catch((error) => {
```
taiyosogawa, line 267: if this succeeds on step `i==0`, but then subsequently fails on step `i==1`, do we need to do anything about the successful `i==0` reconnection?

ts/src/connections/tunnelClientBase.ts
```diff
@@ -236,6 +250,36 @@ export class TunnelClientBase
...
+                // The host is no longer accepting connections on the forwarded port?
+                // Clear the list of disconnected streams for the port, because
+                // it seems it is no longer possible to reconnect them.
+                const streams = this.disconnectedStreams.get(port);
+                if (streams) {
+                    while (streams.length > 0) {
+                        streams.pop()!.dispose();
+                    }
+                }
+            });
```
taiyosogawa, line 282: I noticed the C# implementation of `onForwardedPortAdded` had a success message `"Reconnected stream to forwarded port {port}"`. Should we add that in a `.then()` here too?

ts/src/connections/tunnelClientBase.ts
```diff
@@ -236,6 +250,36 @@ export class TunnelClientBase
...
+        const port = e.port.remotePort;
+        if (typeof port !== 'number') {
+            return;
+        }
+
+        // If there are disconnected streams for the port, re-connect them now.
+        const disconnectedStreamsCount = this.disconnectedStreams.get(port)?.length ?? 0;
+        for (let i = 0; i < disconnectedStreamsCount; i++) {
+            pfs.connectToForwardedPort(port).catch((error) => {
+                this.trace(
```
taiyosogawa, line 268: Is `this.traceWarning()` available here? (Just to make these log lines a little shorter).

ts/src/connections/tunnelClientBase.ts
jasongin: No. Any successfully reconnected streams will be usable. For any that failed to reconnect, the client will have to retry. Note a failure does not break out of the loop -- all the operations are async anyway.

ts/src/connections/tunnelClientBase.ts
jasongin: No. Applications just provide a single callback function for tracing, not an interface.

ts/src/connections/tunnelClientBase.ts
jasongin: Added.

taiyosogawa (dismissed):

taiyosogawa (approved):

PR #288, opened 2023-07-29
Preserve access tokens when refreshing tunnel

jasongin:
When calling management APIs like `getTunnel()` or `updateTunnel()`, if the input Tunnel object has one or more tokens inn its `accessTokens` property, those tokens should be preserved in the returned Tunnel object if not refreshed by the request.

### DIFF
--- a/cs/src/Management/TunnelManagementClient.cs
+++ b/cs/src/Management/TunnelManagementClient.cs
@@ -894,6 +894,7 @@ namespace Microsoft.DevTunnels.Management
                 query: GetApiQuery(),
                 options,
                 cancellation);
+            PreserveAccessTokens(tunnel, result);
             return result;
         }
 
@@ -920,6 +921,7 @@ namespace Microsoft.DevTunnels.Management
                 options,
                 ConvertTunnelForRequest(tunnel),
                 cancellation);
+            PreserveAccessTokens(tunnel, result);
             return result!;
         }
 
@@ -938,14 +940,7 @@ namespace Microsoft.DevTunnels.Management
                 options,
                 ConvertTunnelForRequest(tunnel),
                 cancellation);
-
-            // If no new tokens were requested in the update, preserve any existing
-            // access tokens in the resulting tunnel object.
-            if (options?.TokenScopes == null)
-            {
-                result!.AccessTokens = tunnel.AccessTokens;
-            }
-
+            PreserveAccessTokens(tunnel, result);
             return result!;
         }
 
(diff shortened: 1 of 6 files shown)


cs/src/Management/TunnelManagementClient.cs
```diff
@@ -1335,5 +1325,48 @@ public async Task<NamedRateStatus[]> ListUserLimitsAsync(CancellationToken cance
...
+        /// diagnostics in that case.
+        /// </remarks>
+        private static void PreserveAccessTokens(Tunnel requestTunnel, Tunnel? resultTunnel)
+        {
+            if (requestTunnel.AccessTokens != null && resultTunnel != null)
+            {
+                resultTunnel.AccessTokens ??= new Dictionary<string, string>();
+                foreach (var scopeAndToken in requestTunnel.AccessTokens)
+                {
+                    if (!resultTunnel.AccessTokens.ContainsKey(scopeAndToken.Key))
```
IlyaBiryukov, line 1345: What about complex scope where the key has multiple scopes, should we use `!resultTunnel.TryGetAccessToken(TunnelAccessScope, out _)`

cs/src/Management/TunnelManagementClient.cs
```diff
@@ -1335,5 +1325,48 @@ public async Task<NamedRateStatus[]> ListUserLimitsAsync(CancellationToken cance
...
+        /// This intentionally does not check whether any existing tokens are expired. So
+        /// expired tokens may be preserved also, if not refreshed. This allows for better
+        /// diagnostics in that case.
+        /// </remarks>
+        private static void PreserveAccessTokens(Tunnel requestTunnel, Tunnel? resultTunnel)
+        {
+            if (requestTunnel.AccessTokens != null && resultTunnel != null)
+            {
+                resultTunnel.AccessTokens ??= new Dictionary<string, string>();
+                foreach (var scopeAndToken in requestTunnel.AccessTokens)
```
IlyaBiryukov, line 1343: Would it make sense to check token expiration and do not preserve expired tokens?

IlyaBiryukov:
Please add unit tests.

IlyaBiryukov (approved):

cs/src/Management/TunnelManagementClient.cs
jasongin: See comments on line 1334.

I don't think that would help anything, and it would make diagnosing expired-token problems more difficult.

cs/src/Management/TunnelManagementClient.cs
jasongin: I think they should be distinct. An application could intend to use a single-scope token and multi-scope token in different ways.

IlyaBiryukov (dismissed):

jasongin:
> Please add unit tests.

Done.

IlyaBiryukov (approved):

PR #293, opened 2023-08-23
Custom expiration cli changes

usanth:
Fixes #

### Changes proposed: 
For the new feature `CustomTunnelExpiration` we are adding a property to the Tunnel object that stores the value a user inputs if they want a custom tunnel expiration. This is currently being added as a command line option in both `create` and `host` in tunnels, and these are the associated SDK changes needed to make sure that the value the user specifies in the CLI is reflected when the tunnel is created

### Other Tasks:
- [ ] Added all the associated SDK contracts after adding that new property

### DIFF
--- a/cs/src/Contracts/Tunnel.cs
+++ b/cs/src/Contracts/Tunnel.cs
@@ -138,4 +138,10 @@ public class Tunnel
     /// </summary>
     [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
     public DateTime? Expiration { get; set; }
+
+    /// <summary>
+    /// Gets or the custom amount of time the tunnel will be valid if it is not used or updated.
+    /// </summary>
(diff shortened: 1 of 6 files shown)


cs/src/Contracts/Tunnel.cs
```diff
@@ -138,4 +138,10 @@ public Tunnel()
     /// </summary>
     [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
     public DateTime? Expiration { get; set; }
+
+    /// <summary>
+    /// Gets or the custom amount of time the tunnel will be valid if it is not used or updated.
```
IlyaBiryukov, line 143: What time units? E.g. days, seconds, milliseconds?

usanth: Seconds! Good catch, I can update the comment to reflect that

IlyaBiryukov (approved):

PR #294, opened 2023-08-24
Add program files

usanth:
Fixes #

### Changes proposed: 
- adding the generated program files with the updated contract when I added `CustomExpiration` to the cs contract


IlyaBiryukov:
These files look like they're not really needed. The SDK has no object like 'Program'. It seems to be there is a bug in the generator.

PR #299, opened 2023-08-31
Update Endpoint and add new tunnel contract

jfullerton44:
Fixes #

### Changes proposed: 
- Adds Id property to Endpoints
- Adds TunnelV2 which will be used in the service for the new api-version. In the future we will delete the current tunnel and rename TunnelV2 to Tunnel
-

### Other Tasks:


go/tunnels/tunnel_v.go
```diff
@@ -0,0 +1,71 @@
...
+// Generated from ../../../cs/src/Contracts/TunnelV2.cs
+
+package tunnels
+
+import (
+	"time"
+)
+
+// Data contract for tunnel objects managed through the tunnel service REST API.
+type TunnelV struct {
```
jasongin, line 12: Delete `tunnel_v.go`, `tunnel_v.rs`, `TunnelV.java`, `tunnelV.ts`. These files must have been unintentionally generated while you were working on the C# code, but would not be generated again.

jfullerton44: Removed

cs/src/Contracts/Tunnel.cs
```diff
@@ -37,9 +37,9 @@ public Tunnel()
     /// Gets or sets the generated ID of the tunnel, unique within the cluster.
     /// </summary>
     [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
-    [RegularExpression(TunnelIdPattern)]
-    [StringLength(TunnelIdLength, MinimumLength = TunnelIdLength)]
-    public string? TunnelId { get; set; }
+    [RegularExpression(OldTunnelIdPattern)]
+    [StringLength(OldTunnelIdLength, MinimumLength = OldTunnelIdLength)]
+    public virtual string? TunnelId { get; set; }
```
jasongin, line 42: Is a reason this is made `virtual`?

jasongin (approved):

cs/src/Contracts/Tunnel.cs
jfullerton44: Originally when we were going to share objects it had to be but it is not needed anymore so I removed it

jasongin (approved):

PR #302, opened 2023-09-08
Improve host & client connect options and cancellation

jasongin:
- Introduce `TunnelConnectionOptions` as an optional parameter to the client and host `Connect` API.
   - It allows the application to specify retry and reconnect behavior, and will make it easier to add more connection options in the future without breaking changes.
 - Rename host's `Start` to `Connect` for consistency.
   - `Start` is still available for now, but marked as obsolete/deprecated.
 - Fix and test issues related to retry and cancellation of both C# and TS connections.
   - Add missing cancellation token parameter to TS `connect`.
   - Ensure the correct exception type is thrown if the token is cancelled or the client/host is disposed during the connection or retries.
   - Ensure the client or host instance can be re-used after a failed or cancelled connection attempt.

Above changes apply only to C# and TS code. Other language SDKs may be improved later based on feedback.

### DIFF
--- a/cs/src/Connections/TunnelClient.cs
+++ b/cs/src/Connections/TunnelClient.cs
@@ -632,8 +632,11 @@ public abstract class TunnelClient : TunnelConnection, ITunnelClient
             sshSession.Closed -= OnSshSessionClosed;
         }
 
-        ConnectionStatus = ConnectionStatus.Disconnected;
+        // Clear the SSH session before setting the status to Disconnected, in case the
+        // status-changed event handler immediately triggers annother connection attempt.
         this.SshSession = null;
+
+        ConnectionStatus = ConnectionStatus.Disconnected;
 
         OnSshSessionClosed(e.Exception);
         if (e.Reason == SshDisconnectReason.ConnectionLost)
--- a/cs/src/Connections/TunnelClient.cs
+++ b/cs/src/Connections/TunnelClient.cs
@@ -184,7 +184,7 @@ public abstract class TunnelClient : TunnelConnection, ITunnelClient
             // Unsubscribe event handler from the previous session.
             this.SshSession.Authenticating -= OnSshServerAuthenticating;
             this.SshSession.Disconnected -= OnSshSessionDisconnected;
-            this.SshSession.Closed -= SshSession_Closed;
+            this.SshSession.Closed -= OnSshSessionClosed;
             this.SshSession.Request -= OnRequest;
         }
 
--- a/cs/src/Connections/TunnelClient.cs
+++ b/cs/src/Connections/TunnelClient.cs
@@ -243,11 +243,6 @@ public abstract class TunnelClient : TunnelConnection, ITunnelClient
     private void OnSshSessionDisconnected(object? sender, EventArgs e) =>
         StartReconnectTaskIfNotDisposed();
 
-    private void OnSshSessionClosed(object? sender, EventArgs e)
-    {
-        ConnectionStatus = ConnectionStatus.Disconnected;
-        this.SshSession = null;
-    }
 
     private void ConfigurePortForwardingService()
     {
@@ -508,9 +503,10 @@ public abstract class TunnelClient : TunnelConnection, ITunnelClient
     /// </summary>
     protected virtual void SshSessionCreated()
     {
-        // All tunnel hosts and clients should disable this because they do not use it (for now) and leaving it enabled is a potential security issue.
+        // All tunnel hosts and clients should disable this because they do not use it (for now)
+        // and leaving it enabled is a potential security issue.
         SshPortForwardingService!.AcceptRemoteConnectionsForNonForwardedPorts = false;
-        SshSession!.Closed += SshSession_Closed;
+        SshSession!.Closed += OnSshSessionClosed;
         IsSshSessionActive = true;
     }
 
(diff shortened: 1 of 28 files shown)


cs/src/Connections/TunnelClient.cs
```diff
@@ -240,6 +243,12 @@ protected async Task StartSshSessionAsync(Stream stream, CancellationToken cance
     private void OnSshSessionDisconnected(object? sender, EventArgs e) =>
         StartReconnectTaskIfNotDisposed();
 
+    private void OnSshSessionClosed(object? sender, EventArgs e)
+    {
+        ConnectionStatus = ConnectionStatus.Disconnected;
```
IlyaBiryukov, line 248: Should we unsubscribe the event handler? Otherwise it'll pin the old `SshSession` object to `TunnelClient`

cs/src/Connections/TunnelClient.cs
```diff
@@ -240,6 +243,12 @@ protected async Task StartSshSessionAsync(Stream stream, CancellationToken cance
     private void OnSshSessionDisconnected(object? sender, EventArgs e) =>
         StartReconnectTaskIfNotDisposed();
 
+    private void OnSshSessionClosed(object? sender, EventArgs e)
```
IlyaBiryukov, line 246: How is it different from `SshSession_Closed` ?

cs/src/Connections/TunnelClient.cs
IlyaBiryukov: And other handlers too.

cs/src/Connections/TunnelClient.cs
```diff
@@ -554,7 +563,6 @@ public override async ValueTask DisposeAsync()
         if (this.SshSession != null)
         {
             await this.SshSession.CloseAsync(SshDisconnectReason.ByApplication);
-            this.SshSession.Dispose();
```
IlyaBiryukov, line 557: Why? Is this redundant?

cs/src/Connections/TunnelConnection.cs
```diff
@@ -189,20 +217,43 @@ protected virtual void ValidateAccessToken()
     /// <summary>
     /// Connection status changed event.
     /// </summary>
+    /// <remarks>
+    /// Before any connection attempt is made, the connection status is
```
IlyaBiryukov, line 221: Thanks for adding these docs!

cs/src/Connections/TunnelConnectionOptions.cs
```diff
@@ -0,0 +1,63 @@
...
+    /// Recoverable errors include network connectivity issues, authentication issues (e.g. expired
+    /// access token which may be refreshed before retrying), and service temporarily unavailable
+    /// (HTTP 503). For rate-limiting errors (HTTP 429) only a limited number of retries are
+    /// attempted before giving up.
+    ///
+    /// Retries are performed with exponential backoff, starting with a 100ms delay and doubling
+    /// up to a maximum 12s delay, with further retries using the same max delay.
+    ///
+    /// Note after the initial connection succeeds, the host or client may still become disconnected
+    /// at any time after that. In that case the <see cref="EnableReconnect" /> option controls
```
IlyaBiryukov, line 34: Furthermore, reconnections might be happening about once per week for long-running clients. That's when we do a deployment and Relay pods are restarted.

cs/src/Connections/TunnelRelayTunnelClient.cs
```diff
@@ -116,28 +116,6 @@ protected override Task<ITunnelConnector> CreateTunnelConnectorAsync(Cancellatio
         return Task.FromResult(result);
     }
 
-    /// <summary>
-    /// Connect to the clientRelayUri using accessToken.
-    /// </summary>
-    protected Task ConnectAsync(
```
IlyaBiryukov, line 122: This is used by the Relay service [here](https://devdiv.visualstudio.com/OnlineServices/_git/tunnels?path=/src/Tunnel/Src/DataPlane/Relay/WebForwarding/RelayServiceTunnelClient.cs&version=GBmain&line=127&lineEnd=128&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents). Is there going to be an alternative?

cs/src/Connections/TunnelClient.cs
jasongin: Yes it's redundant. `CloseAsync` [will dispose the session](https://github.com/microsoft/dev-tunnels-ssh/blob/96dfe16e8d4cf1ddbe2dae52560c60899a87d133/src/cs/Ssh/SshSession.cs#L658).

cs/src/Connections/TunnelClient.cs
jasongin: I will combine them.

cs/src/Connections/TunnelRelayTunnelClient.cs
jasongin: It's a mistake, I will restore it.

cs/src/Connections/TunnelConnectionOptions.cs
jasongin: Added comments about that.

cs/src/Connections/TunnelRelayTunnelClient.cs
jasongin: Done.

cs/src/Connections/TunnelClient.cs
jasongin: Done.

cs/src/Connections/TunnelClient.cs
jasongin: Done.

IlyaBiryukov (approved):

IlyaBiryukov (approved):

IlyaBiryukov (approved):

PR #304, opened 2023-09-11
Add extensions needed for ControlPlanManagementClient in relay

jfullerton44:
Since the Relay needs to make calls to the service we need to support some tunnelV2 calls in the SDK for now. These will be deleted when the SDKS moves fully to TunnelV2

### DIFF
--- a/cs/src/Management/TunnelExtensions.cs
+++ b/cs/src/Management/TunnelExtensions.cs
@@ -73,6 +73,64 @@ public static class TunnelExtensions
     }
 
     /// <summary>
+    /// Try to get an access token from <paramref name="tunnel"/> for <paramref name="accessTokenScope"/>.
+    /// </summary>
+    /// <remarks>
+    /// The tokens are searched in <c>Tunnel.AccessTokens</c> dictionary where each
+    /// key may be either a single scope or space-delimited list of scopes.
+    /// </remarks>
+    /// <param name="tunnel">The tunnel to get the access token from.</param>
+    /// <param name="accessTokenScope">Access token scope to get the token for.</param>
+    /// <param name="accessToken">If non-null and non-empty token is found, the token value. <c>null</c> if not found.</param>
+    /// <returns>
+    /// <c>true</c> if <paramref name="tunnel"/> has non-null and non-empty an access token for <paramref name="accessTokenScope"/>;
+    /// <c>false</c> if <paramref name="tunnel"/> has no access token for <paramref name="accessTokenScope"/> or the token is null or empty.
+    /// </returns>
+    /// <exception cref="ArgumentNullException">If <paramref name="tunnel"/> or <paramref name="accessTokenScope"/> is null.</exception>
+    /// <exception cref="ArgumentException">If <paramref name="accessTokenScope"/> is empty.</exception>
+    public static bool TryGetAccessTokenV2(this TunnelV2 tunnel, string accessTokenScope, [NotNullWhen(true)] out string? accessToken)
+    {
+        Requires.NotNull(tunnel, nameof(tunnel));
+        Requires.NotNullOrEmpty(accessTokenScope, nameof(accessTokenScope));
+
+        if (tunnel.AccessTokens?.Count > 0)
+        {
+            var scope = accessTokenScope.AsSpan();
+            foreach (var (key, value) in tunnel.AccessTokens)
+            {
+                // Each key may be either a single scope or space-delimited list of scopes.
+                var index = 0;
+                while (index < key?.Length)
+                {
+                    var spaceIndex = key.IndexOf(' ', index);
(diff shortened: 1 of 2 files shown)


cs/src/Management/TunnelExtensions.cs
```diff
@@ -72,6 +72,64 @@ public static bool TryGetAccessToken(this Tunnel tunnel, string accessTokenScope
...
+    /// <param name="tunnel">The tunnel to get the access token from.</param>
+    /// <param name="accessTokenScope">Access token scope to get the token for.</param>
+    /// <param name="accessToken">If non-null and non-empty token is found, the token value. <c>null</c> if not found.</param>
+    /// <returns>
+    /// <c>true</c> if <paramref name="tunnel"/> has non-null and non-empty an access token for <paramref name="accessTokenScope"/>;
+    /// <c>false</c> if <paramref name="tunnel"/> has no access token for <paramref name="accessTokenScope"/> or the token is null or empty.
+    /// </returns>
+    /// <exception cref="ArgumentNullException">If <paramref name="tunnel"/> or <paramref name="accessTokenScope"/> is null.</exception>
+    /// <exception cref="ArgumentException">If <paramref name="accessTokenScope"/> is empty.</exception>
+    public static bool TryGetAccessTokenV2(this TunnelV2 tunnel, string accessTokenScope, [NotNullWhen(true)] out string? accessToken)
```
IlyaBiryukov, line 91: Can we just add an overload with the same name, `TryGetAccessToken` ? C# will use either the old one or the new one based on the type of `tunnel`.

cs/src/Management/TunnelManagementClient.cs
```diff
@@ -376,6 +376,55 @@ private string UserLimitsPath
...
+        /// reason.</exception>
+        /// <remarks>
+        /// This protected method enables subclasses to support additional tunnel management APIs.
+        /// Authentication will use one of the following, if available, in order of preference:
+        ///   - <see cref="TunnelRequestOptions.AccessToken"/> on <paramref name="options"/>
+        ///   - token provided by the user token callback
+        ///   - token in <paramref name="tunnel"/> <see cref="Tunnel.AccessTokens"/> that matches
+        ///     one of the scopes in <paramref name="accessTokenScopes"/>
+        /// </remarks>
+        protected async Task<TResult?> SendTunnelV2RequestAsync<TRequest, TResult>(
```
IlyaBiryukov, line 411: Can we add an overload with the same name, just different `tunnel` arg type?

cs/src/Management/TunnelManagementClient.cs
```diff
@@ -790,6 +839,46 @@ public void Dispose()
                 options);
         }
 
+        private Uri BuildTunnelV2Uri(
```
IlyaBiryukov, line 842: Same here.

cs/src/Management/TunnelManagementClient.cs
```diff
@@ -825,6 +914,41 @@ public void Dispose()
             return authHeader;
         }
 
+        private async Task<AuthenticationHeaderValue?> GetV2AuthenticationHeaderAsync(
```
IlyaBiryukov, line 917: And same here.

cs/src/Management/TunnelManagementClient.cs
jfullerton44: updated

cs/src/Management/TunnelManagementClient.cs
jfullerton44: updated

cs/src/Management/TunnelManagementClient.cs
jfullerton44: updated

cs/src/Management/TunnelManagementClient.cs
```diff
@@ -458,8 +507,9 @@ private string UserLimitsPath
             where TRequest : class
         {
             var uri = BuildUri(clusterId, path, query, options);
+            Tunnel? tunnel = null;
             var authHeader = await GetAuthenticationHeaderAsync(
-                tunnel: null, accessTokenScopes: null, options);
+                tunnel: tunnel, accessTokenScopes: null, options);
```
IlyaBiryukov, line 512: Why changing from `null` to `tunnel`? It'll change how the auth header is calculated.

IlyaBiryukov: Actually, I see it's `null` anyway.

IlyaBiryukov (approved):

cs/src/Management/TunnelManagementClient.cs
jfullerton44: It makes sure it can differentiate between the functions

PR #307, opened 2023-09-12
Fix connecting to local TCP port in Go SDK

dmgardiner25:
### Changes proposed: 
Uses the TCP listener passed into the `ConnectToForwardedPort` function to actually connect to the port via a local TCP port. This is the same logic that is used by the [GH CLI](https://github.com/cli/cli/blob/c5f88bb551e22b04a3799561f3abea4892012eff/pkg/liveshare/port_forwarder.go#L78C27-L78C44) which is what much of this connection code seems to be based on.

cc. @jfullerton44

### Other Tasks:
- [X] If you updated the Go SDK did you update the PackageVersion in tunnels.go


jfullerton44:
Can you pull the latest version of main and fix conflicts

jfullerton44 (approved):

dmgardiner25:
@jfullerton44 I don't have merge access, mind merging when you get the chance?

PR #309, opened 2023-09-13
Add dash as an allowedIdChar

jfullerton44:
### Changes proposed: 
- Allows new tunnel Ids to contain a dash in any position except the first or last
-
-

### Other Tasks:


cs/src/Contracts/TunnelConstraints.cs
```diff
@@ -195,7 +195,7 @@ public static class TunnelConstraints
     /// excluding vowels and 'y' (to avoid accidentally generating any random words).
     /// </summary>
     /// <seealso cref="Tunnel.TunnelId"/>
-    public const string NewTunnelIdChars = "0123456789abcdefghijklmnopqrstuvwxyz";
+    public const string NewTunnelIdChars = "0123456789abcdefghijklmnopqrstuvwxyz-";
```
JacobRoberts, line 198: Is this being used anymore?

jfullerton44: Not directly, but once we update the tunnel types we can add it to the documentation as it is much easier to understand compared to "[a-z0-9][a-z0-9-]{1,58}[a-z0-9]"

JacobRoberts (approved):

PR #312, opened 2023-09-15
Add TunnelListByRegion and error detail objects

IlyaBiryukov:
* Add `ErrorDetail` class following [Azure API guidelines](https://github.com/microsoft/api-guidelines/blob/vNext/azure/Guidelines.md#http-query-parameters-and-header-values).
* Add `InnerErrorDetail` class
* Handle `[JsonPropertyName]` attribute in generator.
* Add `TunnelListByRegionResponse` and `TunnelListByRegion` contracts for the new list API.
* Box recursive structs in rust.

Fixes for https://github.com/microsoft/basis-planning/issues/1133


cs/src/Contracts/TunnelListByRegion.cs
```diff
@@ -0,0 +1,39 @@
...
+/// <summary>
+/// Tunnel list by region.
+/// </summary>
+public  class TunnelListByRegion
+{
+    /// <summary>
+    /// Azure region name.
+    /// </summary>
+    [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
+    public string? RegionName { get; set; }
```
jfullerton44, line 20: Should we add or reference a list of expected region names to be returned here

IlyaBiryukov: It'll be one of official Azure region names, as listed by `az account list-locations -o table`.
We can list what current regions are in the comments. That may change later when we onboard new regions.

jfullerton44 (dismissed):

jfullerton44 (dismissed):

jfullerton44 (approved):

PR #325, opened 2023-10-03
Allow direct port connection in Go SDK

dmgardiner25:
The main goal of this PR is to unblock the `gh cs ssh --stdio` command in the GitHub CLI. This command redirects stdin/stdout to the stream rather than opening up a local TCP port which was previously supported in Live Share but was not implemented when the Dev Tunnels Go SDK was created. This is done by updating the `ConnectToForwardedPort` to allow for direct connection to an `io.ReadWriteCloser` when connecting to a forwarded port. The previous functionality to pass in a local port listener was migrated to `ConnectListenerToForwardedPort`.

### Changes proposed: 
- Update `ConnectToForwardedPort` to accept an `io.ReadWriteCloser`
- Create `ConnectListenerToForwarded` that accepts a `net.Listener`
- Update tests

### Other Tasks:
- [X] If you updated the Go SDK did you update the PackageVersion in tunnels.go

### DIFF
--- a/go/tunnels/client.go
+++ b/go/tunnels/client.go
@@ -136,35 +136,46 @@ func (c *Client) Connect(ctx context.Context, hostID string) error {
 	return nil
 }
 
-// Opens a stream connected to a remote port for clients which cannot or do not want to forward local TCP ports.
-// Returns a readWriteCloser which can be used to read and write to the remote port.
-// Set AcceptLocalConnectionsForForwardedPorts to false in ConnectAsync to ensure TCP listeners are not created
-// This will return an error if the port is not yet forwarded,
-// the caller should first call WaitForForwardedPort.
-func (c *Client) ConnectToForwardedPort(ctx context.Context, listenerIn *net.TCPListener, port uint16) error {
+// ConnectListenerToForwardedPort opens a stream to a remote port and connects it to a given listener.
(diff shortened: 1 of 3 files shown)


go/tunnels/client.go
```diff
@@ -136,33 +136,26 @@ func (c *Client) Connect(ctx context.Context, hostID string) error {
...
-// the caller should first call WaitForForwardedPort.
-func (c *Client) ConnectToForwardedPort(ctx context.Context, listenerIn *net.TCPListener, port uint16) error {
+// ConnectListenerToForwardedPort opens a stream to a remote port and connects it to a given listener.
+//
+// Ensure that the port is already forwarded before calling this function
+// by calling WaitForForwardedPort. Otherwise, this will return an error.
+//
+// Set acceptLocalConnectionsForForwardedPorts to false when creating the client to ensure
+// TCP listeners are not created for all ports automatically when the client connects.
+func (c *Client) ConnectListenerToForwardedPort(ctx context.Context, listenerIn net.Listener, port uint16) error {
```
dmgardiner25, line 146: Updated the `listenerIn` type to be the generic `net.Listener` in case users don't want to use TCP.

jfullerton44 (approved):

PR #329, opened 2023-10-10
Refactor, fix status change events, reduce 409s, and handle too many connections disconnected host in C# SDK

IlyaBiryukov:
Fix for [#1198](https://github.com/microsoft/basis-planning/issues/1198)
Fix incorrect connection status transitions reported by CS.

* Refactor relay-related common code into `RelayTunnelConnection` class.
* Do not re-update tunnel endpoint in `TunnelRelayTunnelHost` instance for subsequent `ConnectAsync` after the first if they're about the same tunnel endpoint.
* Do not try to delete tunnel endpoint in `TunnelRelayTunnelHost` disposal if the endpoint was not created.
* Deprecate `RefreshintTunnelHostPublicKey` connection status and add `TunnelRefreshing` event to replace it.
* Use common tunnel refreshing code for port refresh.
* Ensure `DisconnectException` is null in connection status changed event args if it's not disconnection.
* Ensure `DisconnectException` property is null if the connection is disposed.
* Ensure session events are unsubscribed when the session is closed and disposed.
* Add `DisconnectReason` property to tunnel relay host and client.
* Do not allow reconnection for tunnel host if it was disconnected with "too many connections" reason. This is to ensure that multiple hosts for the same tunnel do not fight each other.
* Improve disconnection reason tracing.
* Add more unit tests and better test tunnel client and host connection status, disconnect exception and reason, and connection status changed event.
* Increase delay between reconnect attempts after 429, 502, and 503.
* Port changes to ts SDK. Refactor and flatten class hierarchy in ts SDK.


jasongin (dismissed):
This looks great, especially all the additional validation in the tests. It helps ensure applications can rely on consistent connection and reconnection behavior.

I think the TS SDK will need some similar changes. (A separate PR is fine.)

cs/src/Connections/IRelayClient.cs
```diff
@@ -27,8 +27,32 @@ internal interface IRelayClient
...
+    /// <summary>
+    /// Get relay client dispose token.
+    /// </summary>
+    CancellationToken DisposeToken { get; }
+
+    /// <summary>
+    /// Start connecting relay client.
+    /// </summary>
+    /// <exception cref="ObjectDisposedException">If relay client is disposed.</exception>
+    void StartConnecting();
```
jfullerton44, line 44: Does this do anything other than change the connection status?

cs/src/Connections/IRelayClient.cs
```diff
@@ -39,17 +63,25 @@ internal interface IRelayClient
 
     /// <summary>
     /// Configures tunnel SSH session with the given stream.
+    /// If this method succeeds, the SSH session must be connected and ready.
+    /// If this method fails, depending on <see cref="TunnelConnectionOptions.EnableRetry"/> and failure, tunnel client my try reconnecting.
```
jfullerton44, line 67:
```suggestion
    /// If this method fails, depending on <see cref="TunnelConnectionOptions.EnableRetry"/> and failure, tunnel client may try reconnecting.
```

cs/src/Connections/IRelayClient.cs
```diff
@@ -27,8 +27,32 @@ internal interface IRelayClient
...
+    /// <summary>
+    /// Finish connecting relay client.
+    /// </summary>
+    /// <exception cref="ObjectDisposedException">If relay client is disposed.</exception>
+    void FinishConnecting(SshDisconnectReason reason, Exception? disconnectException);
+
     /// <summary>
     /// Create stream to the tunnel.
+    /// If this method succeeds, <see cref="ConfigureSessionAsync(Stream, bool, CancellationToken)"/> will be called.
+    /// If this method fails, depending on <see cref="TunnelConnectionOptions.EnableRetry"/> and failure, tunnel client my try reconnecting.
```
jfullerton44, line 55:
```suggestion
    /// If this method fails, depending on <see cref="TunnelConnectionOptions.EnableRetry"/> and failure, tunnel client may try reconnecting.
```

cs/src/Connections/IRelayClient.cs
```diff
@@ -39,17 +63,25 @@ internal interface IRelayClient
...
     /// </summary>
     Task CloseSessionAsync(SshDisconnectReason disconnectReason, Exception? exception);
 
     /// <summary>
     /// Refresh tunnel access token. This may be useful when the Relay service responds with 401 Unauthorized.
     /// </summary>
+    /// <exception cref="ObjectDisposedException">If the connection is disposed or disconnected.</exception>
+    /// <exception cref="ArgumentException">If current status is not <see cref="ConnectionStatus.Connecting"/>.</exception>
+    /// <exception cref="UnauthorizedAccessException">If the refreshed token is expired.</exception>
+    /// <exception cref="OperationCanceledException">If refresh is cancelled or connection disposed.</exception>
```
jfullerton44, line 84: If connection is disposed how is it determined which of these it will be between ObjectDisposedException and OperationCanceledException

cs/src/Connections/RefreshingTunnelEventArgs.cs
```diff
@@ -0,0 +1,66 @@
...
+
+    /// <summary>
+    /// Get tunnel access scope for this tunnel client or host.
+    /// </summary>
+    public string TunnelAccessScope { get; }
+
+    /// <summary>
+    /// Get the tunnel being refreshed.
+    /// </summary>
+    public Tunnel? Tunnel { get; }
```
jfullerton44, line 44: Why is this nullable, won't there always be a tunnel being refreshed

cs/src/Connections/RelayTunnelConnector.cs
```diff
@@ -48,169 +48,220 @@ public RelayTunnelConnector(IRelayClient relayClient)
...
-            catch (WebSocketException wse)
-            {
-                if (wse.WebSocketErrorCode == WebSocketError.UnsupportedProtocol)
+                    // TODO: check tunnel access token expiration and try refresh it if it's expired.
+                    stream = await this.relayClient.CreateSessionStreamAsync(cts.Token);
+                    await this.relayClient.ConfigureSessionAsync(stream, isReconnect, cts.Token);
+
+                    stream = null;
+                    disconnectReason = SshDisconnectReason.None;
+                    isFinalAttempt = true;
```
jfullerton44, line 77: Is there a reason to set this local variable before returning?

cs/src/Connections/TunnelRelayTunnelHost.cs
```diff
@@ -672,25 +605,20 @@ private void OnSshChannelOpening(object? sender, SshChannelOpeningEventArgs e)
...
             return;
         }
 
-        var updatedTunnel = await ManagementClient.GetTunnelAsync(
-            Tunnel, new TunnelRequestOptions { IncludePorts = true });
-
-        var updatedPorts = updatedTunnel?.Ports ?? Array.Empty<TunnelPort>();
-        Tunnel.Ports = updatedPorts;
-
+        var updatedPorts = Tunnel?.Ports ?? Array.Empty<TunnelPort>();
```
jfullerton44, line 613: If Tunnel.ports is empty or null can we just return immediately

jfullerton44 (dismissed):
Looks good but I'm not a TS expert

cs/src/Connections/IRelayClient.cs
IlyaBiryukov: Yep. This change is needed to isolate where the connection status is set to `Connecting`.

jasongin (approved):

cs/src/Connections/IRelayClient.cs
IlyaBiryukov: It depends on the timing. When relay client is disposed, it cancels its connection operation. If that cancellation is observed first, the method may throw `OperationCancelledException`. If disposal happens a little bit later, it may become `ObjectDisposedException`. Anyway, this is used by `RelayTunnelConnector` where it's then always converted to `ObjectDisposedException` if the relay client is disposed, `OperationCancelledException` if the consumer cancelled connection, or `UnauthorizedAccessException` in other cases.

cs/src/Connections/RelayTunnelConnector.cs
IlyaBiryukov: Yep, it's used in `finally` block that fires on `return` too.

cs/src/Connections/RelayTunnelConnector.cs
```diff
@@ -48,169 +48,220 @@ public RelayTunnelConnector(IRelayClient relayClient)
...
+                                    wse);
+
+                            case HttpStatusCode.NotFound:
+                                throw new TunnelConnectionException($"The tunnel or port is not found (404).", wse);
+
+                            case HttpStatusCode.TooManyRequests:
+                            case HttpStatusCode.ServiceUnavailable:
+                            case HttpStatusCode.BadGateway:
+                                // Normally nginx choses another healthy pod when it cannot establish connection to a pod.
+                                // However, if there are no other pods, it may returns 502 (Bad Gateway) to the client.
```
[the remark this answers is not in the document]
IlyaBiryukov: Neelima is going to change pod behavior to about HTTP connection when nginx need to choose another pod. If nginx cannot choose any, it would return 502. The comment on line 148 explains that.

cs/src/Connections/TunnelRelayTunnelHost.cs
IlyaBiryukov: We also stop forwarding from the ports that are now missing on line 639.

cs/src/Connections/RefreshingTunnelEventArgs.cs
IlyaBiryukov: In some cases relay clients may connect to Relay without the `Tunnel` object. This is used for example by Relay service itself. The tunnel refreshing may kick off without a known tunnel object.

jfullerton44 (approved):

jfullerton44 (approved):

PR #333, opened 2023-10-17
Add default api-version in C#

jfullerton44:
### Changes proposed: 
- Fixes bug where functionality would be broken when api-version was not set

### DIFF
--- a/cs/src/Management/TunnelManagementClient.cs
+++ b/cs/src/Management/TunnelManagementClient.cs
@@ -91,7 +91,7 @@ namespace Microsoft.DevTunnels.Management
         public TunnelManagementClient(
             ProductInfoHeaderValue userAgent,
             Func<Task<AuthenticationHeaderValue?>>? userTokenCallback = null,
-            string? apiVersion = null)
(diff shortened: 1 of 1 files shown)


jasongin (approved):

cs/src/Management/TunnelManagementClient.cs
```diff
@@ -91,7 +91,7 @@ public class TunnelManagementClient : ITunnelManagementClient
         public TunnelManagementClient(
             ProductInfoHeaderValue userAgent,
             Func<Task<AuthenticationHeaderValue?>>? userTokenCallback = null,
-            string? apiVersion = null)
+            string apiVersion = "2023-09-27-preview")
```
usanth, line 94: Would this default version have to get updated every time we update the api version? Thoughts on making the value ` "2023-09-27-preview"` a constant variable/ put it in a centralized place so in the future we just need to update it in one place rather than a couple times in the same file

jfullerton44: good point, I added a constant

usanth (approved):

PR #336, opened 2023-10-18
Add CreateOrUpdate tunnels and ports and forced updates and creates with headers

jfullerton44:
### Changes proposed: 
- Adds sdk functions to create a tunnel or port or update it if it already exists
- Also adds forced creates and updates with If-Match and If-Not-Match headers
-

### Other Tasks:


cs/src/Management/TunnelManagementClient.cs
```diff
@@ -924,6 +925,60 @@ public void Dispose()
...
+                       ConvertTunnelForRequest(tunnel),
+                       cancellation,
+                       true);
+                    PreserveAccessTokens(tunnel, result);
+                    return result!;
+                }
+                catch (UnauthorizedAccessException) when (idGenerated && retries < 3) // The tunnel ID was already taken.
+                {
+                    tunnel.TunnelId = IdGeneration.GenerateTunnelId();
+                }
```
taiyosogawa, line 957: We could consider catching an "expectation failed" exception when `idGenerated == false` to skip additional retries.

cs/src/Management/TunnelManagementClient.cs
```diff
@@ -924,6 +925,60 @@ public void Dispose()
...
+                       path: null,
+                       query: GetApiQuery(),
+                       options,
+                       ConvertTunnelForRequest(tunnel),
+                       cancellation,
+                       true);
+                    PreserveAccessTokens(tunnel, result);
+                    return result!;
+                }
+                catch (UnauthorizedAccessException) when (idGenerated && retries < 3) // The tunnel ID was already taken.
```
taiyosogawa, line 954: Could be worth putting 3 into a constant. Or, change `retries` to `remainingRetries`, initialize to 3, then decrement.

cs/src/Management/TunnelManagementClient.cs
```diff
@@ -1151,6 +1212,51 @@ public void Dispose()
...
+                tunnel,
+                ManagePortsAccessTokenScopes,
+                path,
+                query: GetApiQuery(),
+                options,
+                ConvertTunnelPortForRequest(tunnel, tunnelPort),
+                cancellation))!;
+            PreserveAccessTokens(tunnelPort, result);
+
+            if (tunnel.Ports != null)
```
taiyosogawa, line 1239: If `tunnel.Ports` is null, would we want to initialize it to an empty array, then add the new port(s) instead of skipping?

taiyosogawa: I think the Go implementation will add the ports if the property is initially null, but the java and JS implementation follow the C# pattern.

taiyosogawa (commented):
Nice job getting updates in 4 languages!

cs/src/Management/TunnelManagementClient.cs
jfullerton44: Since this is a "catch when" if idgenerated == false the exception will not be caught and no retries will be done

taiyosogawa: I was thinking about a totally separate branch - basically, if the ID is specified and the creation fails because the ID is taken, it's not worth retrying. It's not an essential piece of logic, and could definitely be optimized later.

jfullerton44: That should be the current behavior. In that case the catch would not happen so the Exception would just be thrown and that would be it

taiyosogawa: Ah, you're right! Thanks!

cs/src/Management/TunnelManagementClient.cs
jfullerton44: update

cs/src/Management/TunnelManagementClient.cs
jfullerton44: Good catch, I made the update

taiyosogawa (approved):

taiyosogawa (approved):

PR #337, opened 2023-10-20
Adding Windows Partner Information to UserAgent C# and Rust Changes

usanth:
Fixes:
C# and Rust SDK changes needed for https://github.com/orgs/microsoft/projects/115/views/18?pane=issue&itemId=39437861

### Changes proposed: 
- Adds operating system and operating system version to the user agent value before making a request
- Adds windows partner id (if it exists) to the user agent value

Changes look like this when creating tunnel in C#:
image

User Agent looks like this when creating tunnel using Rust SDK:
image

### Other Tasks:


jasongin:
> OS/Win32NT
OS-Version/10.0.22621.0
Windows-Partner-Id/e3171dd...
Dev-Tunnels-Service-CSharp-SDK/1.0.7464+c19b72aacb

I realized this does not quite match the usual format for User-Agent headers. To better follow existing conventions, I'd suggest combining these values into one header for the SDK, like this:

`Dev-Tunnels-Service-CSharp-SDK/1.0.7464+c19b72aacb (Microsoft Windows 10.0.22621; PartnerId:e3171dd...)`

I think the part in parentheses is the `ProductInfoHeaderValue.Comment` property.

cs/src/Management/TunnelUserAgent.cs
```diff
@@ -47,4 +49,40 @@ public static class TunnelUserAgent
...
+    /// partner id if applicable
+    /// </summary>
+    /// <returns>List of product info header values with
+    /// machine properties.</returns>
+    public static List<ProductInfoHeaderValue> GetMachineHeaders()
+    {
+        var windowsHeaderValues = new List<ProductInfoHeaderValue>();
+        var os = Environment.OSVersion.Platform;
+        windowsHeaderValues.Add(ProductInfoHeaderValue.Parse("OS" + "/" + os.ToString()));
+        var version = Environment.OSVersion.Version;
```
jasongin, line 65: Use `RuntimeInformation.OSDescription` to get the OS name and version; I think it will be more detailed than `Environment.OSVersion` for non-Windows platforms.

usanth:
>Dev-Tunnels-Service-CSharp-SDK/1.0.7464+c19b72aacb (Microsoft Windows 10.0.22621; PartnerId:e3171dd...)

Rather than doing this, I added another user agent property that contains all this information in a comment. I couldn't append it onto an existing UserAgent value because the `Comment` property is read-only, and there is no constructor of `ProductInfoHeaderValue` that takes 3 constructors. Therefore, it now looks like this:
image

jasongin:
> Rather than doing this, I added another user agent property that contains all this information in a comment.

That works. All values for the same header name are effectively concatenated anyway.

> HTTP: User-Agent: (OS/Microsoft Windows 10.0.22621 Windows-Partner-Id/e3171dd9-9a5f-e5be-b36c-cc7c4f3f3bcf)

The format isn't right for a User-Agent comment: key-value pairs should be separated with a colon, and different fields should be separated with a semicolon. [See the MDN User-Agent doc.](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent)

rs/src/management/http_client.rs
```diff
@@ -12,6 +12,9 @@ use url::Url;
 
 use rand::Rng;
 
+use winreg::enums::*;
+use winreg::RegKey;
```
connor4312, line 16: These have to be in conditional compilation as well, or it'll fail on other platforms. You can put them directly in the block down below.

rs/src/management/http_client.rs
```diff
@@ -533,10 +536,32 @@ pub struct TunnelClientBuilder {
 /// to get the client instance (or cast automatically).
 pub fn new_tunnel_management(user_agent: &str) -> TunnelClientBuilder {
     let pkg_version = PKG_VERSION.unwrap_or("unknown");
-    let full_user_agent = format!(
-        "{}{}{}",
-        user_agent, " Dev-Tunnels-Service-Rust-SDK/", pkg_version
+    let os = os_info::get();
+    let os_info = format!("{}: {} {}", "OS", os.os_type(), os.version());
+    let mut windows_partner_id = String::default();
```
connor4312, line 541: Use an Option type here, set it `= None` initially, set it to `Some(value)` when you have it, and then `if let Some(id) = windows_partner_id` down below

rs/src/management/http_client.rs
```diff
@@ -533,10 +536,32 @@ pub struct TunnelClientBuilder {
...
+    #[cfg(windows)]
+    {
+        let hklm = RegKey::predef(HKEY_LOCAL_MACHINE);
+        let key = hklm.open_subkey("SOFTWARE\\Microsoft\\Windows365");
+        windows_partner_id = key
+            .expect("No value found for key")
```
connor4312, line 547: We shouldn't panic (and prevent usage of consuming applications) if this isn't found imo

rs/src/management/http_client.rs
```diff
@@ -533,10 +536,32 @@ pub struct TunnelClientBuilder {
...
+            .get_value::<String, _>("PartnerId")
+            .unwrap_or_default();
```
connor4312, line 549: once windows_partner_id is an option type, and if we can silently ignore the lack of a key, this can just be `window_partner_id = key.and_then(|k| k.get_value("PartnerId")).ok()`

rs/src/management/http_client.rs
```diff
@@ -533,10 +536,32 @@ pub struct TunnelClientBuilder {
...
+        user_agent, " Dev-Tunnels-Service-Rust-SDK/", pkg_version, os_info
     );
+
+    if windows_partner_id != String::default() {
+        windows_partner_id = format!("Windows-Partner-Id: {}", windows_partner_id);
+        full_user_agent = format!(
+            "{}{}{} ({}; {})",
+            user_agent, " Dev-Tunnels-Service-Rust-SDK/", pkg_version, os_info, windows_partner_id
+        );
+    }
```
connor4312, line 563: nit: I would probably move the user agent builder out to its own function, this is getting a little heavy

rs/Cargo.toml
```diff
@@ -23,6 +23,8 @@ rand = "0.8.5"
 russh = { version = "0.37.1", default-features = false, features = ["openssl", "flate2"], optional = true }
 russh-keys = { version = "0.37.1", default-features = false, features = ["openssl"], optional = true }
 hyper = "0.14"
+os_info = { version = "3", default-features = false }
+winreg = "0.8"
```
connor4312, line 27: This only needs to be a dependency on Windows https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#platform-specific-dependencies

usanth: Added that to the `Cargo.toml` file for just windows

rs/src/management/http_client.rs
usanth: Addressed

rs/src/management/http_client.rs
usanth: Thanks for suggesting this. It also prevents the program from breaking for a lack of key, which I agree it shouldn't do since this key won't exist for many users

connor4312 (approved):
rust lgtm

jasongin (approved):

PR #347, opened 2023-11-03
rs: remove unused mut

mblair:
This just fixes an error I noticed when building:

```
➜  rs git:(main) cargo build
warning: variable does not need to be mutable
   --> src/management/http_client.rs:552:9
    |
552 |     let mut windows_partner_id: Option<String> = None;
    |         ----^^^^^^^^^^^^^^^^^^
    |         |
    |         help: remove this `mut`
    |
    = note: `#[warn(unused_mut)]` on by default

warning: `tunnels` (lib) generated 1 warning (run `cargo fix --lib -p tunnels` to apply 1 suggestion)
    Finished dev [unoptimized + debuginfo] target(s) in 0.14s
```


connor4312 (changes requested):
This is used when building Windows (L559)

PR #348, opened 2023-11-06
Add policies property to tunnelservice request header on C#

neelip:
Fixes # https://github.com/orgs/microsoft/projects/115/views/18?pane=issue&itemId=43806545 

### Changes proposed: 
- Read policies from windows registry and add them to the request header. 
- Updated header:
image

### DIFF
--- a/cs/src/Management/RegistryTools.cs
+++ b/cs/src/Management/RegistryTools.cs
@@ -4,79 +4,95 @@
 // </copyright>
 
 using System;
+using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
+using System.Reflection.PortableExecutable;
 using System.Runtime.InteropServices;
+using System.Security;
 using System.Text;
 using Microsoft.DevTunnels.Contracts;
 using Microsoft.Win32;
 
 namespace Microsoft.DevTunnels.Management
 {
-    internal class RegistryTools
+    /// <summary>
+    /// Provides methods to get and format policy settings from the Windows Registry.
+    /// </summary>
+    public class PolicyProvider
     {
+        private readonly string regKeyPath;
+
         /// <summary>
-        /// Get registry key value from the HKLM root Registry.
+        /// Initializes a new instance of the <see cref="PolicyProvider"/> class.
         /// </summary>
-        /// <param name="regKeyPath"></param>
-        /// <param name="defaultOnError"></param>
-        /// <returns></returns>
-        public object GetRegistryValueFromLocalMachineRoot(string regKeyPath, object? defaultOnError = null)
+        /// <param name="regKeyPath">The registry path where the policy settings are stored on the local machine</param>
+        public PolicyProvider(string regKeyPath)
         {
-            object regValue = "";
-            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
-            {
-                RegistryKey rootKey = RegistryKey.OpenBaseKey(RegistryHive.LocalMachine, RegistryView.Registry32);
-                regValue = this.GetRegistryValue(rootKey, regKeyPath, defaultOnError);
-            }
-            return regValue;
+            this.regKeyPath = regKeyPath ?? throw new ArgumentException(nameof(regKeyPath));
         }
 
         /// <summary>
-        /// Get registry key settings int value.
+        /// Get formatted registry values as a header string from the Local Machine Hive. 
         /// </summary>
-        /// <param name="rootKey">Root key entry</param>
-        /// <param name="regKeyPath">Path to Key</param>
-        /// <param name="defaultOnError"></param>
-        /// <returns></returns>
-        private object GetRegistryValue(RegistryKey rootKey, string regKeyPath, object? defaultOnError = null)
+        /// <param name="defaultOnError">default value to return on error</param>
+        /// <returns>A string formatted for use in an Http header, or a 'null' as a default value.</returns>
+        public string? GetHeaderValue(string? defaultOnError = null)
         {
-            StringBuilder headerBuilder = new StringBuilder();
-
+            
+            string regValue = string.Empty;
+            if (!RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                return defaultOnError;
+            }
             try
             {
-                if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
-                {
-                    if (rootKey != null)
-                    {
-                        using (RegistryKey? subKey = rootKey.OpenSubKey(regKeyPath, RegistryKeyPermissionCheck.ReadSubTree, System.Security.AccessControl.RegistryRights.QueryValues))
-                        {
-                            if (subKey != null)
-                            {
-                                foreach (string valueName in subKey.GetValueNames())
-                                {
-                                    object? value = subKey.GetValue(valueName);
-                                    if (value != null)
-                                    {
-                                        string headerValue = value.ToString()!;
+                using var rootKey = RegistryKey.OpenBaseKey(RegistryHive.LocalMachine, RegistryView.Registry32);
+                regValue = this.CreateHeaderString(rootKey, regKeyPath);
 
-                                        headerBuilder.AppendFormat("{0}={1}; ", Uri.EscapeDataString(valueName), Uri.EscapeDataString(headerValue!));
-                                    }
-                                }
-                            }
-                        }
-
-                        if (headerBuilder.Length > 0)
-                        {
-                            headerBuilder.Length -= 2;
-                        }
-                    }
-                }
             }
             catch
             {
+                return defaultOnError;
             }
+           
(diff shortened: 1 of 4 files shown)


cs/src/Management/RegistryTools.cs
```diff
@@ -0,0 +1,79 @@
+// <copyright file="TunnelExtensions.cs" company="Microsoft">
```
nandiniYeltiwar, line 1: nit: copyright file name is wrong.

cs/src/Management/RegistryTools.cs
```diff
@@ -0,0 +1,79 @@
...
+                    if (rootKey != null)
+                    {
+                        using (RegistryKey? subKey = rootKey.OpenSubKey(regKeyPath, RegistryKeyPermissionCheck.ReadSubTree, System.Security.AccessControl.RegistryRights.QueryValues))
+                        {
+                            if (subKey != null)
+                            {
+                                foreach (string valueName in subKey.GetValueNames())
+                                {
+                                    object? value = subKey.GetValue(valueName);
+                                    string headerValue = value != null ? value.ToString()! : "Not Set";
```
nandiniYeltiwar, line 58: If "value == null", lets not add the registry to the header. Or maybe just pass empty string.

cs/src/Management/RegistryTools.cs
```diff
@@ -0,0 +1,79 @@
...
+        }
+
+        /// <summary>
+        /// Get registry key settings int value.
+        /// </summary>
+        /// <param name="rootKey">Root key entry</param>
+        /// <param name="regKeyPath">Path to Key</param>
+        /// <param name="defaultOnError"></param>
+        /// <returns></returns>
+        private object GetRegistryValue(RegistryKey rootKey, string regKeyPath, object? defaultOnError = null)
```
nandiniYeltiwar, line 41: This method is not returning  RegistryValue but a header string. Lets rename the method accordingly.

cs/src/Management/RegistryTools.cs
```diff
@@ -0,0 +1,82 @@
...
+using Microsoft.Win32;
+
+namespace Microsoft.DevTunnels.Management
+{
+    internal class RegistryTools
+    {
+        /// <summary>
+        /// Get registry key value from the HKLM root Registry.
+        /// </summary>
+        /// <param name="regKeyPath"></param>
```
IlyaBiryukov, line 20: Nit: If the xml docs are empty, just remove them.

cs/src/Management/TunnelManagementClient.cs
```diff
@@ -519,6 +519,15 @@ private string UserLimitsPath
 
             request.Headers.UserAgent.Add(TunnelSdkUserAgent);
 
+            const string regKeyPath = @"Software\Policies\Microsoft\Tunnels";
+            var defaultOnError = true;
+            RegistryTools policies = new RegistryTools();
+            var keyValue = (string)policies.GetRegistryValueFromLocalMachineRoot(regKeyPath, defaultOnError);
```
nandiniYeltiwar, line 525: We can extract this piece of code in a separate method such that there is a policyProvider object that returns list of policies and policy values. PolicyProvider class can have the registry path set and intialize RegistryTools and it calls getregistryValue that you have specified in another file.

cs/src/Management/RegistryTools.cs
```diff
@@ -0,0 +1,82 @@
...
+                        if (headerBuilder.Length > 0)
+                        {
+                            headerBuilder.Length -= 2;
+                        }
+                    }
+                }
+            }
+            catch
+            {
+            }
```
IlyaBiryukov, line 77: Should we trace the exception in `catch` block?

cs/src/Management/RegistryTools.cs
```diff
@@ -0,0 +1,82 @@
...
+                                    {
+                                        string headerValue = value.ToString()!;
+
+                                        headerBuilder.AppendFormat("{0}={1}; ", Uri.EscapeDataString(valueName), Uri.EscapeDataString(headerValue!));
+                                    }
+                                }
+                            }
+                        }
+
+                        if (headerBuilder.Length > 0)
```
IlyaBiryukov, line 68: Do you mean `if (headerBuilder.Length >= 2)` ?

cs/src/Management/RegistryTools.cs
```diff
@@ -0,0 +1,82 @@
...
+        /// <param name="regKeyPath">Path to Key</param>
+        /// <param name="defaultOnError"></param>
+        /// <returns></returns>
+        private object GetRegistryValue(RegistryKey rootKey, string regKeyPath, object? defaultOnError = null)
+        {
+            StringBuilder headerBuilder = new StringBuilder();
+
+            try
+            {
+                if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
```
IlyaBiryukov, line 47: Nit: you can simplify nested ifs if you check for negative condition and bail out eary:
``` csharp
 if (!RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
{
    return string.Empty;
}
```

cs/src/Management/RegistryTools.cs
```diff
@@ -0,0 +1,82 @@
...
+        }
+
+        /// <summary>
+        /// Get registry key settings int value.
+        /// </summary>
+        /// <param name="rootKey">Root key entry</param>
+        /// <param name="regKeyPath">Path to Key</param>
+        /// <param name="defaultOnError"></param>
+        /// <returns></returns>
+        private object GetRegistryValue(RegistryKey rootKey, string regKeyPath, object? defaultOnError = null)
```
IlyaBiryukov, line 41: Should we check if `rootKey != null` and throw? `Requires.NotNull(rootKey, nameof(rootKey))`. Otherwise, what's the point of the method if does nothing with null `rootKey` ?

cs/src/Management/RegistryTools.cs
```diff
@@ -0,0 +1,82 @@
...
+{
+    internal class RegistryTools
+    {
+        /// <summary>
+        /// Get registry key value from the HKLM root Registry.
+        /// </summary>
+        /// <param name="regKeyPath"></param>
+        /// <param name="defaultOnError"></param>
+        /// <returns></returns>
+        public object GetRegistryValueFromLocalMachineRoot(string regKeyPath, object? defaultOnError = null)
```
IlyaBiryukov, line 23: Check if `regKeyPath` is not null.

cs/src/Management/RegistryTools.cs
```diff
@@ -0,0 +1,82 @@
...
+        {
+            object regValue = "";
```
IlyaBiryukov, line 25: Nit: use `string.Empty` instead of ""

cs/src/Management/TunnelManagementClient.cs
```diff
@@ -519,6 +519,15 @@ private string UserLimitsPath
...
+            if (keyValue != null)
```
IlyaBiryukov, line 526: It'll always be not null. We return empty string if it's not windows or the registry key is missing.

cs/src/Management/TunnelManagementClient.cs
```diff
@@ -519,6 +519,15 @@ private string UserLimitsPath
 
             request.Headers.UserAgent.Add(TunnelSdkUserAgent);
 
+            const string regKeyPath = @"Software\Policies\Microsoft\Tunnels";
+            var defaultOnError = true;
+            RegistryTools policies = new RegistryTools();
+            var keyValue = (string)policies.GetRegistryValueFromLocalMachineRoot(regKeyPath, defaultOnError);
+            if (keyValue != null)
+            {
+                request.Headers.Add("Policies", keyValue);
```
IlyaBiryukov, line 528: How about `User-Agent-Policies` header name if it's not decided yet?

Also can we limit the length of `keyValue` to some reasonable length? Our ingress has max 64K header length, but that's for all of the headers. So, how about limiting this header to 1K?

neelip: In this case should Policies be added as to the existing User-Agent property?

cs/src/Management/RegistryTools.cs
neelip: It is expected that if the `headerBuilder` is not empty then it has a key-value pair appended in the "{0}={1}; " format.
I am not sure why the check would be specifically for greater than 2 characters?

IlyaBiryukov (approved):

cs/src/Management/TunnelManagementClient.cs
IlyaBiryukov: Having policies in the User-Agent header is ok. There would be more work to parse it from there on the server side or in telemetry queries. Having policies in its own header is also ok. Less work for the service and telemetry queries to parse it. I'm just thinking that we already have policies. So, to not confuse them with these policies, let's name the header differently, e.g. `User-Agent-Policies`.

nandiniYeltiwar (approved):

PR #349, opened 2023-11-08
Add policies property to tunnelservice request header on Rust

neelip:
Fixes # https://github.com/orgs/microsoft/projects/115/views/18?pane=issue&itemId=43806545

Changes proposed:
Read policies from windows registry and add them to the request header.
Updated header:
image


rs/src/management/http_client.rs
```diff
@@ -456,6 +460,12 @@ impl TunnelManagementClient {
         let headers = request.headers_mut();
         headers.insert("User-Agent", self.user_agent.clone());
 
+        // Add Windows group policies to the header
+        let policy_header_value = self.policy_provider.get_header_value("").await?;
+        if !policy_header_value.is_empty() {
```
connor4312, line 465: You should use the Option type here instead of treating empty strings specially. Then you can do something like `if let Some(h) = policy_header_value { /* h is a String here */ }`

rs/src/management/http_client.rs
```diff
@@ -456,6 +460,12 @@ impl TunnelManagementClient {
...
+            headers.insert("Policies", HeaderValue::from_str(&policy_header_value)?);
```
connor4312, line 466: nit: I think you can use [`HeaderValue::from_maybe_shared`](https://docs.rs/http/latest/http/header/struct.HeaderValue.html) to avoid copying memory here

rs/src/management/http_client.rs
```diff
@@ -456,6 +460,12 @@ impl TunnelManagementClient {
         let headers = request.headers_mut();
         headers.insert("User-Agent", self.user_agent.clone());
 
+        // Add Windows group policies to the header
+        let policy_header_value = self.policy_provider.get_header_value("").await?;
```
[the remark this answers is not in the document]
neelip: In c# when this function fails to return value it would throw an exception. I have modified the code to handle the error.

rs/src/management/policy_provider.rs
```diff
@@ -0,0 +1,37 @@
+use std::io;
+
+pub const REGISTRY_KEY_PATH: &str = r"Software\Policies\Microsoft\Tunnels";
+
+#[cfg(target_os = "windows")]
+pub fn get_policy_header_value(default_on_error: &str) -> io::Result<Option<String>> {
```
connor4312, line 6: default_on_error seems unused now

rs/src/management/mod.rs
```diff
@@ -3,9 +3,11 @@ mod errors;
...
 mod tunnel_locator;
 mod tunnel_request_options;
+mod policy_provider;
 
 pub use authorization::*;
 pub use errors::*;
 pub use http_client::*;
 pub use tunnel_locator::*;
 pub use tunnel_request_options::*;
+pub use policy_provider::*;
```
connor4312, line 13: `pub use` exposes the function to consumers of the crate. I think this is an implementation detail that doesn't need to be public

rs/src/management/http_client.rs
```diff
@@ -456,6 +459,21 @@ impl TunnelManagementClient {
         let headers = request.headers_mut();
         headers.insert("User-Agent", self.user_agent.clone());
 
+        // Add Windows group policies to the header
+        match get_policy_header_value("None") {
+            Ok(Some(policy_header_value)) => {
+                let header_value = HeaderValue::from_maybe_shared(policy_header_value)
+                    .expect("Invalid header value");
```
connor4312, line 466: Maybe this should be an error log as well?

connor4312 (commented):
Looking good overall, a few final tweaks. I would also run `cargo fmt` and `cargo clippy` to format and check for lint errors

connor4312 (approved):

PR #366, opened 2023-12-07
TLS Cert Exception

usanth:
Fixes #https://github.com/devdiv-microsoft/basis-planning/issues/1328

### Changes proposed: 
- Catching the certificate exception in the C# SDK and displaying a better error message to the user
- Catching the exception in the TS SDK. The error that gets returned is of type Error, so we parse the error message to see if it is certificate related and provide a more helpful exception message 

### Other Tasks:


cs/src/Management/TunnelManagementClient.cs
```diff
@@ -549,12 +551,24 @@ private string UserLimitsPath
...
+                response,
+                cancellation);
+                return result;
+            }
+            catch (HttpRequestException ex) when (ex.InnerException is AuthenticationException auex)
+            {
+                throw new HttpRequestException($"Error: Tunnel service HTTPS certificate is invalid. This may" +
+                    $" be caused by the use of a firewall intercepting the connection.", auex);
+            }
+            catch (Exception ex)
```
jasongin, line 568: Is this catch clause needed? The exception may already be of type `HttpRequestException`, and in that case re-wrapping it loses some details. And otherwise I'm not sure it is good to change the type of exception.

ts/src/management/tunnelManagementHttpClient.ts
```diff
@@ -672,9 +672,17 @@ export class TunnelManagementHttpClient implements TunnelManagementClient {
...
-        const result = await this.request<TResult>(method, uri, body, config, allowNotFound);
-        this.raiseReportProgress(TunnelProgress.CompletedSendTunnelRequest);
-        return result;
+        try {
+            const result = await this.request<TResult>(method, uri, body, config, allowNotFound);
+            this.raiseReportProgress(TunnelProgress.CompletedSendTunnelRequest);
+            return result;
+        } catch (error) {
+            if (/certificate/i.test((error as Error).message)) {
+                throw new Error("Tunnel service HTTPS certificate is invalid. This may be caused by the use of a "+
```
jasongin, line 681: The original error message should be included also, since it might contain useful information.

ts/src/management/tunnelManagementHttpClient.ts
jasongin, line 1: When updating the TS library, now we need to update the versions in `package.json` also. See [the comment in this other PR](https://github.com/microsoft/dev-tunnels/pull/382/files/2098079ba42de229e355592dccd8bb7bce82dc63#diff-8d3585007ad625cd833b9d27d8637dcb53fdd553652d22d5905c1c508fe4e5ed).

jasongin (dismissed):

cs/src/Management/TunnelManagementClient.cs
usanth: Exception messages are read-only. The only way that I found to be able to alter the exception message was to create a new exception and use the older exception as the inner exception. This is the same exception with the same inner exception, so minimal details should be lost

ts/src/management/tunnelManagementHttpClient.ts
usanth: Included orignal message at the end of the new error message

ts/src/management/tunnelManagementHttpClient.ts
```diff
@@ -672,9 +672,18 @@ export class TunnelManagementHttpClient implements TunnelManagementClient {
...
-        return result;
+        try {
+            const result = await this.request<TResult>(method, uri, body, config, allowNotFound);
+            this.raiseReportProgress(TunnelProgress.CompletedSendTunnelRequest);
+            return result;
+        } catch (error) {
+            if (/certificate/i.test((error as AxiosError<any>).message)) {
+                const originalErrorMessage = (error as AxiosError<any>).message;
+                throw new Error("Tunnel service HTTPS certificate is invalid. This may be caused by the use of a " +
+                    "self-signed certificate or a firewall intercepting the connection." + originalErrorMessage + ". ");
```
derekbekoe, line 683:
```suggestion
                    "self-signed certificate or a firewall intercepting the connection. " + originalErrorMessage + ".");
```

Small nit on formatting to do with the spaces.
As is currently in the PR, I believe you may end up with an error message such as `...or a firewall intercepting the connection.Axios Error Message. `.

ts/src/management/tunnelManagementHttpClient.ts
```diff
@@ -703,9 +712,17 @@ export class TunnelManagementHttpClient implements TunnelManagementClient {
...
-        this.raiseReportProgress(TunnelProgress.CompletedSendTunnelRequest);
-        return result;
+        try {
+            const result = await this.request<TResult>(method, uri, body, config, allowNotFound);
+            this.raiseReportProgress(TunnelProgress.CompletedSendTunnelRequest);
+            return result;
+        } catch (error) {
+            if (/certificate/i.test((error as Error).message)) {
+                throw new Error("Tunnel service HTTPS certificate is invalid. This may be caused by the use of a "+
+                "self signed certificate or a firewall intercepting the connection.");
```
derekbekoe, line 722: Is there a need to include the original error message here also?

usanth: When I run this locally my exception message is just "self signed certificate". I think that would not be descriptive enough to add as an addition to this exception message, but if it would mean consistency between the ts and cs files then I can add it

ts/src/management/tunnelManagementHttpClient.ts
usanth: Addressed

cs/src/Management/TunnelManagementClient.cs
jasongin: My comment above is referring to these lines:
```C#
catch (Exception ex)
{
    throw new HttpRequestException(ex.Message);
}
```

This re-throw is not changing the exception message.

When we get to this `catch` block, either the exception is not of type `HttpRequestException` or it that type but without the expected inner exception type. In either case it is not certificate-related so there is no need to handle it. Wrapping it in `HttpRequestException` does not provide any benefit.

derekbekoe (dismissed):

cs/src/Management/TunnelManagementClient.cs
usanth: Good catch, initially with the comment I thought you were referring to the other catch block.

This makes sense, I went ahead and removed this catch block entirely since the exception would just have been rethrown anyways.

jasongin (approved):

jasongin (approved):

jasongin (approved):

PR #369, opened 2023-12-11
Update Nerdbank Versioning in TS and CS

jfullerton44:
* add version files

* update version

* remove other refs

* more update refs

* temp remove

Fixes #

### Changes proposed: 
-
-
-

### Other Tasks:


jasongin (commented):
I don't understand why we need the same `version.json` file in the `cs` and `ts` subdirectories. Does it fix the problem of the versions being out of sync? And then why isn't it needed in the `go`, `java`, `rs` subdirectories?

jasongin (commented):
Should we also update the version of the `NerdBank.GitVersioning` nuget and npm packages that are referenced? I think they are pretty old now.

jfullerton44:
> I don't understand why we need the same `version.json` file in the `cs` and `ts` subdirectories. Does it fix the problem of the versions being out of sync? And then why isn't it needed in the `go`, `java`, `rs` subdirectories?

It seems like this thing that fixed the issue was including ../cs in the ts folder and ../ts in the cs folder. That meant that the version would go up for each repo whenever a change was made to either one. Before it looks like it was just incrementing for its own changes. Unfortunately, we cant have ../cs or ../ts in the base version.json file as this throws an error as the path from that point goes outside the repo. For go, java, and rust, I believe the versions are chosen manually, so nerdbank does not apply there, so they do not need a version.json file

jasongin:
> Unfortunately, we cant have ../cs or ../ts in the base version.json file as this throws an error as the path from that point goes outside the repo.

The path filter is [relative to the version.json file](https://github.com/dotnet/Nerdbank.GitVersioning/blob/main/doc/pathFilters.md#path-filter-format), so it would not make sense to have ../cs or ../ts in the root version.json.

The reason the versions are out of sync is that we [previously had a version.json file under the `cs` directory](https://github.com/microsoft/dev-tunnels/commits/main/cs/version.json). So when NBGV calculates the [git commit height](https://github.com/dotnet/Nerdbank.GitVersioning/blob/main/README.md#what-is-git-height) it finds a shorter commit path to a version.json file in the `cs` directory.

I think the simplest/best way to get the versions in sync again is to increment the minor version in the root version.json file.

> Git 'height' is the number of commits in the longest path from HEAD to the commit that set the major.minor version number to the values found in HEAD.

Then all subfolders will find the height to the commit of the root version.json file where the minor version was last incremented. (We can also reset the offset.)

jfullerton44:
Updated to just modify the global version file

klvnraju (approved):

PR #370, opened 2023-12-12
Allow client app to cancel fwd of specific ports

jasongin:
Normally a tunnel client will forward all ports requested by the host. However in some scenarios it's useful for a client to be able to selectively decline to forward some port numbers. This change adds a `TunnelClient.PortForwarding` event with a `Cancel` property. The event notifies the client application when a port is about to be forwarded as requested by the host and allow cancellation of that request.

### DIFF
--- a/cs/src/Connections/ITunnelClient.cs
+++ b/cs/src/Connections/ITunnelClient.cs
@@ -27,6 +27,20 @@ public interface ITunnelClient : IAsyncDisposable
     IReadOnlyCollection<TunnelConnectionMode> ConnectionModes { get; }
 
     /// <summary>
+    /// Event raised when a port is about to be forwarded to the client.
+    /// </summary>
+    /// <remarks>
+    /// The application may cancel this event to prevent specific port(s) from being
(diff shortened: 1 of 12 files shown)


cs/src/Connections/ITunnelClient.cs
```diff
@@ -26,6 +26,20 @@ public interface ITunnelClient : IAsyncDisposable
     /// </summary>
     IReadOnlyCollection<TunnelConnectionMode> ConnectionModes { get; }
 
+    /// <summary>
```
IlyaBiryukov, line 29: What if `WaitForForwardedPortAsync` is called for the port that was cancelled?

IlyaBiryukov (approved):

cs/src/Connections/ITunnelClient.cs
jasongin: It will continue waiting until the port is forwarded (again) without being cancelled, which I think is correct behavior. If the client application both waits for and cancels the same port number, that is their bug.

PR #372, opened 2023-12-15
Add connection report progress infrastructure

jramsay:
Fixes internal #1254

Add connection report progress infrastructure for TS

Note: support in other languages to follow

### Changes proposed: 
- update to consume latest ssh ssh-tcp packages that include reportProgress
- add reportProgress event 
- add connnection events

### Other Tasks:


ts/src/connections/tunnelRelayTunnelHost.ts
```diff
@@ -462,6 +463,7 @@ export class TunnelRelayTunnelHost extends TunnelConnectionSession implements Tu
             config.addService(PortForwardingService);
         });
         session.trace = this.trace;
+        session.onReportProgress(this.raiseReportProgress, this, this.sshSessionDisposables);
```
IlyaBiryukov, line 466: Is this OK if progress is reported from multiple SSH sessions? The main one that connects sdk host to Relay, and different client sessions? The progress sequence may be quite chaotic.

jramsay: I think so. The progress events aren't very verbose. Really only tracks 6 steps to help diagnose connection issues

IlyaBiryukov: But there will be repetitive and out-of-order steps coming from different sessions.

jramsay: thanks for pointing this case out! As discussed, I'll switch to using event args and add a clientSessionId that can be used to determine which client session is reporting the event.

IlyaBiryukov (approved):

PR #373, opened 2023-12-18
Add enum for Api Version in C# and TS

jfullerton44:
Fixes #

### Changes proposed: 
- Adds an enum for api versions in C# and TS SDKs
-
-

### Other Tasks:

### DIFF
--- a/ts/src/management/tunnelManagementHttpClient.ts
+++ b/ts/src/management/tunnelManagementHttpClient.ts
@@ -38,7 +38,7 @@ const clustersApiPath = '/clusters';
 const tunnelAuthentication = 'Authorization';
 const checkAvailablePath = ':checkNameAvailability';
 const createNameRetries = 3;
-enum ManagementApiVersions {
+export enum ManagementApiVersions {
     Version20230927preview = '2023-09-27-preview',
 }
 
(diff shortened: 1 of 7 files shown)


ts/src/management/tunnelManagementHttpClient.ts
```diff
@@ -38,6 +38,9 @@ const clustersApiPath = '/clusters';
 const tunnelAuthentication = 'Authorization';
 const checkAvailablePath = ':checkNameAvailability';
 const createNameRetries = 3;
+enum ManagementApiVersions {
```
connor4312, line 41: This should probably be exported so consumers can use the enum value instead of having to use the string

jfullerton44: Updated to export it

ts/src/management/tunnelManagementHttpClient.ts
```diff
@@ -133,7 +136,7 @@ export class TunnelManagementHttpClient implements TunnelManagementClient {
      */
     public constructor(
         userAgents: (ProductHeaderValue | string)[] | ProductHeaderValue | string,
-        apiVersion: string,
+        apiVersion: ManagementApiVersions,
```
jramsay, line 139: Perhaps add an `apiVersion` param to the constructor signature comments above for consistency.

jramsay (approved):

ts/src/management/tunnelManagementHttpClient.ts
jfullerton44: good catch

jramsay (approved):

PR #358, opened 2023-12-04
Require new packages for TS interdependencies

jfullerton44:
Fixes #

### Changes proposed: 
- Requires that TS packages use the newest version of the other packages when imported. This fixes the issue where yarn can import multiple version of the contracts packages

PR #382, opened 2024-01-22
Add cancellation token and default timeout (20s) to service requests

jramsay:
Fixes #

### Changes proposed: 
- add cancellation token to all tunnelManagementClient requests
- add default connection/service response timeouts
- add unit tests

### Other Tasks:

- [✅ ] If you updated the TS SDK did you update the dependencies in package.json for connections and management to require a dependency that is > the current published version(Found using `npm view @microsoft/dev-tunnels-contracts`). This will fix issues where yarn will pull the old version of packages and will cause mismatched dependencies. See [example PR](https://github.com/microsoft/dev-tunnels/pull/358)


osortega:
Would it be better to handle request retries completely within the SDK if there is already a way to track a request done through the SSH session?
This way it would be a transparent implementation on the user side

ts/src/connections/tunnelRelayTunnelClient.ts
```diff
@@ -499,15 +499,15 @@ export class TunnelRelayTunnelClient extends TunnelConnectionSession implements
...
+    public async refreshPorts(cancellation?: CancellationToken): Promise<void> {
         if (!this.sshSession || this.sshSession.isClosed) {
             throw new Error('Not connected.');
         }
 
         const request = new SessionRequestMessage();
         request.requestType = 'RefreshPorts';
         request.wantReply = true;
-        await this.sshSession.request(request);
+        await this.sshSession.request(request, cancellation);
```
osortega, line 510: How long would a request take to get successfully canceled? I'm just wondering if we go with this approach, how long would the user have to wait before retrying.

jasongin: Cancelling an SSH session request causes the SSH session to close, because it can interrupt sending/receiving protocol messages and leave the protocol in an unknown state. The cancellation parameter for the `SshSession.request()` method is meant to be used only when shutting down the session. So you can't cancel and retry a session request like this.

I don't think there is much value in retrying the request from here anyway: the SSH protocol guarantees the message will be received by the other side (unless the entire session fails). The right place to add retry would be on the [host's handling of that request](https://github.com/microsoft/dev-tunnels/blob/869adcec1c69b73a011a407a7ea8143e9ad54f79/ts/src/connections/tunnelRelayTunnelHost.ts#L632), when it calls the service to request updated ports info.

jasongin (changes requested):

ts/src/connections/tunnelRelayTunnelClient.ts
osortega: What would mean to cancel the `refreshPorts` request from the host side (which ultimately calls `refreshTunnel`)? Would that just cancel the service request and the SSH session will remain intact? Or would that require the client to re-establish the SSH connection?

jasongin: Yes on the host side cancelling would just cancel the service request and leave any client SSH sessions intact. (After the response from the service it sends notification messages about any changes over SSH to connected clients, but that part can't be cancelled.)

ts/src/connections/package.json
```diff
@@ -18,8 +18,8 @@
 		"buffer": "^5.2.1",
 		"debug": "^4.1.1",
 		"vscode-jsonrpc": "^4.0.0",
-		"@microsoft/dev-tunnels-contracts": ">1.1.1",
-		"@microsoft/dev-tunnels-management": ">1.1.1",
+		"@microsoft/dev-tunnels-contracts": ">1.1.6",
+		"@microsoft/dev-tunnels-management": ">1.1.6",
```
jasongin, line 22: Unrelated, but do we have to update these versions with every change now? If so it would be better if it was automatic as part of the build/publish process.

ts/src/connections/tunnelClient.ts
```diff
@@ -104,6 +104,8 @@ export interface TunnelClient extends TunnelConnection {
      * removed ports will then propagate back to the set of ports forwarded by the current
      * client. After the returned task has completed, any newly added ports are usable from
      * the current client.
+     *
+     * @param cancellation Optional cancellation for the request.
      */
-    refreshPorts(): Promise<void>;
+    refreshPorts(cancellation?: CancellationToken): Promise<void>;
```
jasongin, line 110: Revert this change also.

jasongin (approved):

ts/src/connections/package.json
jramsay: ya, it is listed as part of the PR comments now when you create a new SDK PR:

> If you updated the TS SDK did you update the dependencies in package.json for connections and management to require a dependency that is > the current published version(Found using npm view @microsoft/dev-tunnels-contracts). This will fix issues where yarn will pull the old version of packages and will cause mismatched dependencies. 

I can create an item for the build/publish task though as that is a good idea

ts/src/connections/tunnelRelayTunnelClient.ts
jramsay: thanks! Will add retry logic if we see that it is necessary after the next round of results.

PR #385, opened 2024-01-25
Updating to the latest version of axios

jramsay:
### Changes proposed: 
- updating to the latest version of axios
- update to consistent error message prefix for 'ECONNABORTED' scenarios
- return 'ECONNABORTED' error.code for both cases to allow for consumers to retry in this scenario

### Other Tasks:

- [X] If you updated the TS SDK did you update the dependencies in package.json for connections and management to require a dependency that is > the current published version(Found using `npm view @microsoft/dev-tunnels-contracts`). This will fix issues where yarn will pull the old version of packages and will cause mismatched dependencies. See [example PR](https://github.com/microsoft/dev-tunnels/pull/358)


setaskin (approved):
LGTM

PR #396, opened 2024-02-13
Update typescript-build-and-publish.yaml for Azure Pipelines

jfullerton44:
The changes were generated from the ESRP extension in ADO


neelip (approved):
nit: maybe a little bit more detail in the description

PR #405, opened 2024-02-22
Update displayed error message for enterprise policies

neelip:
Fixes [#54175005](https://github.com/orgs/devdiv-microsoft/projects/68/views/3?pane=issue&itemId=54175005)

### Changes proposed: 
- Currently the response shows as such: 
image
- Updated to only show the text in "detail" rathe than displaying the whole json object in the response. 

### Other Tasks:

- [ ] If you updated the TS SDK did you update the dependencies in package.json for connections and management to require a dependency that is > the current published version(Found using `npm view @microsoft/dev-tunnels-contracts`). This will fix issues where yarn will pull the old version of packages and will cause mismatched dependencies. See [example PR](https://github.com/microsoft/dev-tunnels/pull/358)


cs/src/Management/TunnelManagementClient.cs
```diff
@@ -1,1566 +1,1570 @@
...
+                {
+                    case HttpStatusCode.BadRequest:
+                        throw new ArgumentException(errorMessage, hrex);
+
+                    case HttpStatusCode.Unauthorized:
+                    case HttpStatusCode.Forbidden:
+                        // Enterprise Policies
+                        if (response.Headers.Contains("X-Enterprise-Policy-Failure"))
+                        {
+                            var message = response.Content != null ? await response.Content.ReadAsStringAsync() : string.Empty;
```
IlyaBiryukov, line 689: Does the dev tunnels service already send out JSON error details?

IlyaBiryukov: As the `ErrorDetails` class is changing in the same PR, I assume not yet. Then will this code break when it encounters a string response from the service where it expects JSON?  
If the content type of the response will change, it can be used to distinguish between a string or JSON. Or wrap `JsonSerializer.Deserialize<ErrorDetails>(message)` in `try-catch` and if it fails, assume it's a string, not json.

IlyaBiryukov (commented):
Is this going to be a breaking change for the client because the service still returns a string, not JSON in this case?

cs/src/Management/TunnelManagementClient.cs
neelip: Yes, the service already sends out json error details with the new response header key `X-Enterprise-Policy-Failure`.

neelip: The suggestion on the service side was to send the response in ProblemDetails rather than a string value. All 3 policies send ProblemDetails json object instead of a string. And given this specific parsing will happen only when Enterprise policy response header is present, the expected content should be in json format. I don't see an exception to this, unless the response for a new enterprise policy in the future is send in a different format.

cs/src/Management/TunnelManagementClient.cs
```diff
@@ -729,10 +732,11 @@ private string UserLimitsPath
...
         /// </remarks>
-        private class ErrorDetails
-        {
-            public string? Message { get; set; }
-            public string? StackTrace { get; set; }
+        private class ErrorDetails
+        {
+            public string? Message { get; set; }
+            public string? StackTrace { get; set; }
+            public string? detail { get; set; }
```
IlyaBiryukov, line 739: Why `detail` starts with lowercase but `Message` and `StackTrace` are capitalized? Shouldn't `JsonSerializer` handle the lowercase translation (maybe with an additional attribute or serializer settings)?

neelip:
> Is this going to be a breaking change for the client because the service still returns a string, not JSON in this case?

The Enterprise policies should be sending the json object with the latest release. All 3 policies only send JSON object.

cs/src/Management/TunnelManagementClient.cs
neelip: I added an additional check to make sure there is no exception if the string is not in a json format.

IlyaBiryukov (dismissed):

usanth (approved):

PR #411, opened 2024-03-04
Disambiguate `tunnel connection is disposed` error

jramsay:
Disambiguate `tunnel connection is disposed` error

- [x ] If you updated the TS SDK did you update the dependencies in package.json for connections and management to require a dependency that is > the current published version(Found using `npm view @microsoft/dev-tunnels-contracts`). This will fix issues where yarn will pull the old version of packages and will cause mismatched dependencies. See [example PR](https://github.com/microsoft/dev-tunnels/pull/358)


ts/src/connections/tunnelConnectionBase.ts
```diff
@@ -168,9 +168,11 @@ export class TunnelConnectionBase implements TunnelConnection {
     /**
      * Throws CancellationError if the tunnel connection is disposed.
      */
-    protected throwIfDisposed() {
+    protected throwIfDisposed(message: string, stack: string|undefined = undefined) {
```
IlyaBiryukov, line 171: just `stack?: string` should do.

ts/src/connections/tunnelConnectionBase.ts
```diff
@@ -168,9 +168,11 @@ export class TunnelConnectionBase implements TunnelConnection {
...
         if (this.isDisposed) {
-            throw new ObjectDisposedError('The tunnel connection is disposed.');
+            const error = new ObjectDisposedError(`The tunnel connection is disposed. ${message}`);
+            error.stack = stack;
```
IlyaBiryukov, line 174: Only set it if `stack` is defined? Otherwise, can it wipe out existing `stack` on `error` object?

IlyaBiryukov (approved):

IlyaBiryukov (approved):

PR #418, opened 2024-03-08
Detect 4xx error response from firewall

jasongin:
When handling a 4xx error response from a tunnel management API request, check for the absence `X-Served-By` response header to detect when the error was likely caused by a firewall intercepting the request.

### DIFF
--- a/cs/src/Management/TunnelManagementClient.cs
+++ b/cs/src/Management/TunnelManagementClient.cs
@@ -29,6 +29,7 @@ namespace Microsoft.DevTunnels.Management
     /// </summary>
     public class TunnelManagementClient : ITunnelManagementClient
     {
+        private const string ServedByHeaderPrefix = "tunnels-";
         private const string ApiV1Path = "/api/v1";
         private const string TunnelsV1ApiPath = ApiV1Path + "/tunnels";
         private const string SubjectsV1ApiPath = ApiV1Path + "/subjects";
@@ -679,6 +680,22 @@ namespace Microsoft.DevTunnels.Management
                     // A default error message will be filled in below.
                     innerException = ex;
                 }
+            }
+
+            if (errorMessage == null &&
+                (int)response.StatusCode >= 400 && (int)response.StatusCode < 500 &&
+                (!response.Headers.TryGetValues("X-Served-By", out var servedBy) ||
+                 !servedBy.First().StartsWith(ServedByHeaderPrefix)))
+            {
+                // The response did not include either a ProblemDetails body object or a header
+                // confirming it was served by the tunnel service. This check excludes 5xx status
+                // responses which may include non-firwall network infrastructure issues.
+                var requestUri = response.RequestMessage?.RequestUri ??
+                    new Uri(TunnelServiceProperties.Production.ServiceUri);
+                errorMessage = "The tunnel request resulted in " +
+                    $"{(int)response.StatusCode} {response.StatusCode} status, but the request " +
+                    "did not reach the tunnel service. This may indicate the domain " +
+                    $"'{requestUri.Host}' is blocked by a firewall.";
             }
 
             errorMessage ??= "Tunnel service response status code: " + response.StatusCode;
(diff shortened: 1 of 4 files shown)


jasongin:
I'm pretty sure this is not going to work in browsers, since the `X-Served-By` header will be hidden by CORS. I need to test that. And then we might change CORS configuration on the service to [expose](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Expose-Headers) it.

jasongin:
This needs to wait on the CORS change in the service that adds `X-Served-By` to exposed headers.

jasongin:
Reviewers: There appears to be a line-endings difference with `TunnelManagementClient.cs`; I suggest enabling the "Hide whitespace" option when reviewing.

image

cs/src/Management/TunnelManagementClient.cs
```diff
@@ -644,936 +645,952 @@ private string UserLimitsPath
...
+                catch (Exception ex)
+                {
+                    // A default error message will be filled in below.
+                    innerException = ex;
+                }
+            }
+
+            if (errorMessage == null &&
+                (int)response.StatusCode >= 400 && (int)response.StatusCode < 500 &&
+                (!response.Headers.TryGetValues("X-Served-By", out var servedBy) ||
```
usanth, line 687: Just for my own understanding, what does the X-Served-By header have to do with detecting a firewall? Is it just assumed if the returned status code is a 4xx and that this header values indicates it couldn't reach the service then we can assume that it was due to a firewall?

jasongin: The tunnel service always includes the header `X-Served-By: tunnels-cluster-....` in every response. If we get a response that doesn't have that header, or has a different value), that means the response was not from the tunnel service. So the only other possibility is that the response was from a proxy or firewall that intercepted and blocked the request.

ts/src/management/tunnelManagementHttpClient.ts
```diff
@@ -817,8 +817,27 @@ export class TunnelManagementHttpClient implements TunnelManagementClient {
...
+                errorMessage = 'The tunnel request resulted in ' +
+                    `${error.response.status} status, but the request ` +
+                    'did not reach the tunnel service. This may indicate the domain ' +
+                    `'${requestDomain}' is blocked by a firewall.`;
+            }
+        }
+
         if (!errorMessage) {
-            if (error?.response) {
+            if (error.response) {
```
usanth, line 840: Is removing the null ? intentional here? Are we sure in this case that the response is not null?

jramsay (dismissed):

jramsay:
note: still testing before merging

ts/src/management/tunnelManagementHttpClient.ts
usanth: This makes sense, thanks!

ts/src/management/tunnelManagementHttpClient.ts
```diff
@@ -817,8 +817,27 @@ export class TunnelManagementHttpClient implements TunnelManagementClient {
             }
         }
 
+        if (!errorMessage && error.response && error.response.status &&
+            error.response.status >= 400 && error.response.status < 500 &&
+            error.response.headers
+        ) {
+            const headers = error.response.headers;
+            const servedBy = headers['X-Served-By'] || headers['x-served-by'];
+            if (!/tunnel-/.test(servedBy)) {
```
jramsay, line 826: I think this might need to be:

/tunnel**s**-/

jramsay: Tested from the browser using these changes in Live Share. Before making the request I changed the authorization header to an invalid entry. This results in a 401 and hits this code path. The `x-served-by` header is as follows:  

image

jramsay: Results in the following errorMessage:
"The tunnel request resulted in 401 status, but the request did not reach the tunnel service. This may indicate the domain 'global.rel.tunnels.api.visualstudio.com' is blocked by a firewall."

jramsay: I can see you have it correct in the C# SDK. Wil make this change in TS.

usanth (approved):

PR #486, opened 2024-09-27
Add .NET 8 target, remove .NET Core 3.1

jasongin:
- Update from .NET 6 to .NET 8 SDK in `global.json`
 - Add `net8.0` target framework.
 - Remove `netcoreapp3.1` target framework, because .NET Core 3.1 has been unsupported for a couple years now.
 - Update code coverage reporting for .NET 8 compatibility.
 - Update websocket HTTP error handling to use new API available in .NET 8.


cs/global.json
```diff
@@ -1,5 +1,5 @@
 {
   "sdk": {
-    "version": "6.0.412"
+    "version": "8.0.400"
```
IlyaBiryukov, line 3: Add `"rollForward": "latestFeature"` ?

jasongin: Good idea, done.

cs/src/Directory.Build.targets
```diff
@@ -6,9 +6,9 @@
     <Target
         Name="GenerateDocumentation"
         AfterTargets="Build"
-        Condition="'$(GenerateDocs)'=='true' AND '$(TargetFramework)' == 'netcoreapp3.1'"
+        Condition="'$(GenerateDocs)'=='true' AND '$(TargetFramework)' == 'net6.0'"
     >
-        <!-- This target only runs for the .NET 3.1 target; xmldocmd tool currently does not work on .NET 6. -->
+        <!-- This target only runs for the .NET 6 target; xmldocmd does not yet support .NET 8. -->
```
IlyaBiryukov, line 11: `8` or `6`?

jasongin: The comment is correct, this tool doesn't support .NET 8 yet. So it is conditional on `net6.0` target.

I think we don't actually generate these docs anymore though; they were used for the internal docs site.

jramsay (approved):

PR #493, opened 2024-10-17
Re-encode tenant unicode characters in rust SDK

jfullerton44:
Fixes https://github.com/devdiv-microsoft/basis-planning/issues/1658

### Changes proposed: 
- Before sending the User-Agent-Policies header, we need to re-encode the header value to ensure that it is correctly formatted. This manually reencodes the " and ; as they are the only accepted characters in this field. Old value looked like %22tenantId%3DtenantId%22
- Adds formatting changes

### DIFF
--- a/rs/src/management/http_client.rs
+++ b/rs/src/management/http_client.rs
@@ -162,7 +162,11 @@ impl TunnelManagementClient {
     ) -> HttpResult<TunnelEndpoint> {
         let mut url = self.build_tunnel_uri(
             locator,
-            Some(&format!("{}/{}", ENDPOINTS_API_SUB_PATH, endpoint.id.as_deref().unwrap())),
+            Some(&format!(
+                "{}/{}",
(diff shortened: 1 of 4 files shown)


rs/src/management/http_client.rs
```diff
@@ -458,7 +466,16 @@ impl TunnelManagementClient {
         match get_policy_header_value() {
             Ok(Some(policy_header_value)) => {
                 if let Ok(header_value) = HeaderValue::from_maybe_shared(policy_header_value) {
-                    headers.insert("User-Agent-Policies", header_value);
+                    let header_value_str = header_value
+                        .to_str()
+                        .unwrap()
+                        .replace("%22", "")
+                        .replace("%3B", ";");
```
connor4312, line 473: We already use the urlencoding module in this crate, you should be able to use it here: https://docs.rs/urlencoding/latest/urlencoding/#

PR #572, opened 2026-03-02
Adding test pipeline for Symbols publishing AAD auth

plequere-ms:
Fixes #

### Changes proposed: 

Dummy/temporary pipeline to test Symbols publishing with AAD auth. Unfortunately, it needs to be in main as Azure DevOps does not supporting creating pipelines from branches.


ArunGopinathan (approved):
approved.

PR #619, opened 2026-04-22
Split non-prod GitHub App into separate per-environment apps and new Local environment config

jpablo2002:
# Fixes

Splits the shared NonProdGitHubAppClientId into three separate GitHub OAuth apps with dedicated client IDs:

 - PPE (Iv1.b231c327f1eaa229) — existing app, unchanged
 - Dev (Iv23ctTiak9wLCiTcEbr) — new app for the deployed dev service
 - Local (Iv23cttBYzKThF88PiPR) — new app with localhost callback URLs for local development

### Changes proposed: 

Adds a new Local environment to TunnelServiceProperties across all 5 SDK languages (C#, TypeScript, Go, Java, Rust),
with a "local" case in the Environment() method.

This enables the tunnel service and CLI to use different GitHub apps per environment, so each app can have properly
scoped callback URLs and independent secret rotation.


cs/src/Contracts/TunnelServiceProperties.cs
```diff
@@ -137,7 +155,20 @@ private TunnelServiceProperties(
...
+
+    /// <summary>
+    /// Gets properties for the service when running locally.
+    /// </summary>
+    /// <remarks>
+    /// Uses the same service app IDs as the development environment, but a different
+    /// GitHub app with localhost callback URLs.
+    /// </remarks>
+    public static TunnelServiceProperties Local { get; } = new TunnelServiceProperties(
+        $"https://{DevDnsName}/",
```
klvnraju: Should this be DevDnsName for local?

jpablo2002: It should be fine as is. This is how it used to be for local when it used the Development TunnelServiceProperties. The ServiceUri is a default fallback; in practice, local consumers always override it with an explicit localhost URL (e.g. we override it via DEVTUNNELS_SERVICE=local in the CLI)

jpablo2002: I've pushed new changes to add the LocalDnsName to  be tunnels.local.api.visualstudio.com:9901

jpablo2002:
Git diffs and generation of files broke and got messy, abandoning and making new PR

PR #625, opened 2026-04-24
Update DEV environment 3PA app IDs for new Tunnels 3PA DEV app in Test Tenant

jpablo2002:
### Changes proposed: 
The Tunnels1st party app previously used for the DEV environment was disabled, as we're no longer allowed to have 1PA in the Microsoft tenant. The previously created 3PA in the Microsoft tenant is not allowed to become multitenant to allow the E2E tests to use it. Therefore, I've created a new 3PA in our test tenant, which can be multitenant, to allow the previously reenabled DEV usage, and the E2E tests to use it.

This PR pdates the DEV service 3rd-party app (3PA) application IDs to the new values across all five SDK languages (C#, Go, Java, Rust, TypeScript). Changes are limited to TunnelServiceProperties-related files — both manually maintained and auto-generated — and are purely value replacements to use the new 3PA in DEV.

### Other Tasks:

- [x] If you updated the TS SDK did you update the dependencies in package.json for connections and management to require a dependency that is > the current published version(Found using `npm view @microsoft/dev-tunnels-contracts`). This will fix issues where yarn will pull the old version of packages and will cause mismatched dependencies. See [example PR](https://github.com/microsoft/dev-tunnels/pull/358)

### DIFF
--- a/cs/src/Contracts/TunnelServiceProperties.cs
+++ b/cs/src/Contracts/TunnelServiceProperties.cs
@@ -51,14 +51,14 @@ public class TunnelServiceProperties
     internal const string PpeFirstPartyAppId = "54c45752-bacd-424a-b928-652f3eca2b18";
 
     /// <summary>
-    /// Third-party app ID: `DEV-VSTunnelService-3P`
+    /// Third-party app ID for the DEV service environment.
     /// </summary>
     /// <remarks>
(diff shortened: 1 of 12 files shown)


leti367 (approved):
LGTM

adamasmar (approved):

klvnraju (approved):

PR #624, opened 2026-04-24
rust: update dependencies

connor4312:
⚠️ Please do a traditional merge commit to avoid conflicts https://github.com/microsoft/dev-tunnels/pull/626

- chrono->jiff as chrono is going into maintence mode https://github.com/chronotope/chrono/issues/1423#issuecomment-3846148832
- async-trait replaced when possible with Rust 1.75+ native async traits; kept for compat with russh
- futures->futures_util, underlying library for just the things we need
- urlencoding->percent-encoding as it's already a dep of the url module, so just one less dependency
- general version updates

Verified in the VS Code CLI that tunnel functionality still works as expected with these updates.

PR #626, opened 2026-04-27
rust: implement relay client

connor4312:
📣  Builds upon #624 -- only the last commit in this PR is new

Rust had a relay host, but never actually implemented a relay client.

- Imnplement relay_tunnel_client.rs (of course)
- Updated incorrect type generation of some Rust models. These were just never hit in my code paths before, previously list_tunnels/get_tunnels could never return TunnelRelayTunnelEndpoint information. Now our Rust structure matches what Go does where TunnelRelayTunnelEndpoint is a field inside the TunnelEndpoint struct, so its fields are accessible.
- GPT 5.5 did some very impressive analysis to find the bug that led to adding `CHANNEL_WRITE_CHUNK_SIZE`. This was just never something I hit consistently enough before to repro and fix.

### DIFF
--- a/rs/src/connections/relay_tunnel_client.rs
+++ b/rs/src/connections/relay_tunnel_client.rs
@@ -1,0 +1,386 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+use std::{
+    env, io,
+    net::{IpAddr, Ipv4Addr, SocketAddr},
+    pin::Pin,
+    sync::Arc,
+    task::Poll,
+    time::Duration,
+};
+
+use crate::{contracts::TunnelEndpoint, management::TunnelManagementClient};
+use async_trait::async_trait;
+use russh::ChannelMsg;
+use tokio::{
+    io::{AsyncRead, AsyncReadExt, AsyncWrite, AsyncWriteExt},
+    net::TcpListener,
+    sync::watch,
+};
+
+use super::{
+    errors::TunnelError,
+    ws::{build_websocket_request, connect_directly, connect_via_proxy, AsyncRWWebSocket},
+};
+
+/// The RelayTunnelClient connects to a tunnel as a client via the tunneling
+/// service's relay. After connecting, you can open connections to individual
+/// forwarded ports via `connect_to_port`, or forward them to local TCP
+/// listeners via `forward_port_locally`.
+///
+/// # Interoperability
+///
+/// This client uses SSH `direct-tcpip` channels to connect to forwarded ports.
+/// This works with Rust and TypeScript/C# tunnel hosts. Go tunnel hosts do
+/// not currently handle `direct-tcpip` channels and will reject connections
+/// from this client.
+///
+/// # Example
+///
+/// ```ignore
+/// let mgmt = TunnelManagementClient::new(/* ... */);
+/// let client = RelayTunnelClient::new(mgmt);
+///
+/// // endpoint and token are obtained from the tunnel management API
+/// let handle = client.connect(&endpoint, "access_token").await?;
+///
+/// // Connect to port 8080 on the tunnel host
+/// let channel = handle.connect_to_port(8080).await?;
+///
+/// // Or forward a remote port to a local TCP listener
+/// let local_addr = handle.forward_port_locally(8080).await?;
+/// ```
+pub struct RelayTunnelClient {
+    pub proxy: Option<String>,
+    mgmt: TunnelManagementClient,
+}
+
+impl RelayTunnelClient {
+    pub fn new(mgmt: TunnelManagementClient) -> Self {
+        RelayTunnelClient {
+            proxy: env::var("HTTPS_PROXY").or(env::var("https_proxy")).ok(),
+            mgmt,
+        }
+    }
+
+    /// Connects to the tunnel relay as a client. The endpoint must have a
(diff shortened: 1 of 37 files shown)


rs/src/connections/relay_tunnel_client.rs
```diff
@@ -0,0 +1,386 @@
...
+
+/// A connection to a forwarded port on the tunnel host. This is the
+/// client-side equivalent of `ForwardedPortConnection`.
+pub struct PortConnection {
+    channel: russh::Channel<russh::client::Msg>,
+}
+
+impl PortConnection {
+    /// Sends data on the connection.
+    pub async fn send(&mut self, d: &[u8]) -> Result<(), ()> {
```
DavidObando: It looks like this function loses useful context for callers (e.g. log lines like "channel closed while writing on port {}"). Returning `Result<(), russh::Error>` (or `io::Error`) would be more debuggable. Same shape exists on the host side, it looks consistent and I'd be okay if we defer fixing this in a separate PR.

rs/src/connections/relay_tunnel_client.rs
```diff
@@ -0,0 +1,386 @@
...
+                        break;
+                    }
+                }
+                Err(e) => {
+                    log::debug!("error reading from local TCP on port {}: {}", port, e);
+                    conn.close().await;
+                    break;
+                }
+            },
+            data = conn.recv() => match data {
```
DavidObando, line 271: On conn.recv() == None this breaks. The local TCP stream would presumably then be dropped, which RST/closes both halves rather than shutting down the write half first. For protocols sensitive to half-close it can drop in-flight bytes. Consider `stream.shutdown().await` before breaking.

rs/src/connections/relay_tunnel_client.rs
```diff
@@ -0,0 +1,386 @@
...
+    loop {
+        tokio::select! {
+            n = stream.read(&mut read_buf) => match n {
+                Ok(0) => {
+                    log::debug!("EOF from local TCP stream on port {}", port);
+                    conn.close().await;
+                    break;
+                }
+                Ok(n) => {
+                    if conn.send(&read_buf[..n]).await.is_err() {
```
DavidObando, line 259: This code branch breaks without closing conn on is_err(), which looks inconsistent with the Ok(0) branch.

rs/src/contracts/tunnel_endpoint.rs
DavidObando, line 19: Because of the `#[serde(flatten)]` changes TunnelEndpoint always serializes hostRelayUri: null, clientRelayUri: null, etc. when unset, because the generator doesn't add skip_serializing_if = "Option::is_none" for Option<T> (only for Vec). The host send-path now relies on the service tolerating those nulls. Worth a quick sanity check against the service / TS host.

rs/src/connections/relay_tunnel_client.rs
```diff
@@ -0,0 +1,386 @@
...
+impl russh::client::Handler for TunnelClientHandler {
+    type Error = russh::Error;
+
+    async fn check_server_key(
+        self,
+        _server_public_key: &russh_keys::key::PublicKey,
+    ) -> Result<(Self, bool), Self::Error> {
+        // The relay authenticates via the access token; we don't need to
+        // verify the host key.
+        Ok((self, true))
```
DavidObando: This has a code smell to me, but I'm not an expert on the client semantics. @jasongin can you take a look at this `check_server_key` function?

rs/src/connections/relay_tunnel_client.rs
```diff
@@ -0,0 +1,386 @@
...
+        Ok(PortConnection { channel })
+    }
+
+    /// Opens a local TCP listener that forwards connections to the specified
+    /// port on the tunnel host. Returns the local address the listener is
+    /// bound to (useful when the requested port is already in use and a
+    /// fallback port was chosen).
+    ///
+    /// The listener will be stopped when `close()` is called on this handle.
+    pub async fn forward_port_locally(&self, port: u16) -> Result<SocketAddr, TunnelError> {
```
DavidObando, line 183: Passing `&self` instead of `&mut self` means a caller can register the same port twice; second bind will fail and fall back to a different port, likely by mistake and possibly surprising. Consider tracking forwarded ports.

DavidObando (commented):
Consulting on one security-sensitive item. Will approve if things check out.

rs/src/connections/relay_tunnel_client.rs
connor4312: Taking `&mut self` wouldn't prevent that -- that's just a logic error from the caller. But yea we should return an error instead of warning.

rs/src/contracts/tunnel_endpoint.rs
connor4312: It looks like Go adds `omitempty` for all optional types. I'll do the same for Rust generation.

rs/src/connections/relay_tunnel_client.rs
connor4312: Done. Oh how much easier this is than when I first wrote this code...

DavidObando (approved):

connor4312:
thanks for the approval! I do not have write access, so I can't merge this myself. Feel free to merge if you think it's good 🙂

DavidObando:
@connor4312 this has been merged.

PR #672, opened 2026-08-25
Switch ESRP service connection to wif2 to fix npm publish

DavidObando:
## Problem

`typescript-build-and-publish.yaml` fails in the `EsrpRelease` task, so **no npm package is published**. The task fails while acquiring a token, before the release gateway is ever contacted:

```
UseManagedIdentity: true
using managed identity for esrp release authentication for 142047f4-...
Error while acquiring the token :-
invalid_client: Error(s): 700213 - AADSTS700213: No matching federated identity
record found for presented assertion subject '***'. Check your federated identity
credential Subject, Audience and Issuer against the presented assertion.
```

The `Devtunnels-esrp-ame-msi` service connection's federated identity credential no longer matches the assertion subject presented by the ADO v5 agent.

## Why it surfaced now

This pipeline is path-triggered on `ts`, so it runs only when the TypeScript sources change. It last ran successfully on **2026-06-22**; the next run was **2026-08-25**, which failed. The service connection broke somewhere in between and nothing exercised it until now.

`cs-ci-build.yaml` is unaffected — it is a separate pipeline, triggered on `cs`, that pushes through the `dev-tunnels-nuget` feed credential and never touches ESRP. As a result the NuGet packages have continued to publish while the npm packages have fallen behind.

## Fix

`dev-tunnels-ssh` hit this exact failure and resolved it by switching to the `Devtunnels-esrp-ame-msi-wif2` connection in microsoft/dev-tunnels-ssh@d8a9ff1 ("Switch ESRP service connection to wif2 for ADO agent v5 compatibility"). Its pipeline has published successfully ever since — most recently in [build 15078613](https://devdiv.visualstudio.com/OnlineServices/_build/results?buildId=15078613&view=results), where the same `EsrpRelease@11` task returned a Release Id.

This applies the equivalent one-line change here.

Everything else about the ESRP configuration is deliberately unchanged and already matches the working SSH pipeline:

| Setting | Value |
| --- | --- |
| Tenant | `33e01921-4d64-4f8c-a055-5bdaffd5e33d` |
| Client id | `142047f4-eda8-4853-8776-c2e81803ea13` |
| Key Vault | `tunnels-ppe-esrp-kv` |
| Signing certificate | `esrp-sign` |
| Task version | `EsrpRelease@11` |

The surrounding comment block is updated to record the reason, so the next person to read it does not restore the old connection.

## Before merging

Please confirm that **`Devtunnels-esrp-ame-msi-wif2` is authorized for this pipeline** (`Dev-Tunnels SDK TypeScript CI Build`, definition `16353`). Service connections require per-pipeline authorization unless they are open to all pipelines, so the first run may otherwise stop on an authorization prompt.

## Validation

The YAML parses, and `EsrpConnectedServiceName` resolves through to the task's `connectedservicename` input. The change cannot be exercised further without running the release pipeline itself, which publishes.


ArunGopinathan (approved):
approve

